PR #10, opened 2016-03-15
require scipy>=0.14

peterstangl:
flavio uses the `allow_singular` argument of `scipy.stats.multivariate_normal` which is newly introduced in scipy version 0.15.


DavidMStraub:
Thanks! In fact it would be better to make  everything compatible with scipy 0.14 as otherwise too many people will be forced to upgrade scipy. The `allow_singular` is anyway just a hack, so if this is the only place 0.15 is required, it would be better to remove it.

peterstangl:
After removing the `allow_singular` argument from the `scipy.stats.multivariate_normal.logpdf` function, all nosetests finish without errors with scipy 0.14.

DavidMStraub:
Awesome.

PR #18, opened 2016-09-22
Lhcb measurements

DavidMStraub:
Thanks @sreicher for all the additions! I made the following changes:
- I temporarily commented out the A_i measurements in Bs->phimumu as the angular CP asymmetries in Bs->phimumu are not yet implemented (it is a bit more complicated than B->K*mumu due to the non-negligible width difference between Bs and Bsbar.) (b18fcc8)
- I removed the total branching ratios of B->K*mumu and Bs->phimumu that come from an interpolation over the whole q^2 range as these are difficult to define on the theory side (46cdc58)

Also, so far the `measurements.yml` always contained only the most recent measurement of a given quantity by any given experiment, so that new measurements "replace" the old ones, at least if they contain overlapping data sets. This is because in fits one doesn't want to include both. So at the moment I removed the ones that have been superseded by newer ones (73d08dc). On the other hand, it's of course great to have all of them to see how constraints evolve with time with new experimental measurements. The question is where to put them. One option would be to provide a separate repository with a YAML file containing all the superseded measurements. What do you think?


sreicher:
For completeness I would add a seperate repository. Maybe it would be useful to provide several files split by channel or groups of measurements?

DavidMStraub:
OK I will open a new repository. Yes, splitting by channels and maybe even years might be useful.

PR #20, opened 2016-10-10
[physics][bll] Added effective lifetimes for bll.

ChristophNiehoff:
Added effective lifetimes for the decay B_s -> l+ l-.

### DIFF
--- a/flavio/physics/bdecays/bll.py
+++ b/flavio/physics/bdecays/bll.py
@@ -123,6 +123,36 @@ def bqll_obs_function(function, B, l1, l2):
     return lambda wc_obj, par: bqll_obs(function, wc_obj, par, B, l1, l2)
 
 
+# Bs -> l+l- effective lifetime
+
+def tau_ll(wc, par, B, lep):
+    r"""Effective B->l+l- lifetime as defined in eq. (26) of arXiv:1204.1737 .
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Great, thanks!

Can you please add some simple unit test to `test_bll.py` to keep the coverage up? Can be something trivial like reproducing (28) of 1204.1737.

flavio/physics/bdecays/bll.py
```diff
@@ -141,6 +171,18 @@ def bqll_obs_function(function, B, l1, l2):
     _obs.tex = r"$\text{BR}(B^0\to "+_tex[l]+"^+"+_tex[l]+"^-)$."
     Prediction(_obs_name, bqll_obs_function(br_inst, 'B0', l, l))
 
+    # Add the effective lifetimes for Bs
+    _obs_name = 'tau_'+l+l
+    _obs = Observable(_obs_name)
+    _obs.set_description(r"Effective lifetime for $B_s \to "+_tex[l]+"^+"+_tex[l]+"^-$.")
+    _obs.tex = r"$\tau_{B_s \to " +_tex[l] +_tex[l] + "}$."
+    if l=='e':
```
ChristophNiehoff: No, I tried it first. But since python does not give the value of 'l', but rather a reference on it, all lifetimes will effectively be 'tau_tautau' in the end. (Because 'tau' is the last value l takes...)

PR #21, opened 2016-10-13
[plots] Add basic functionality for band plot caching.

apuignav:
Now `band_plot` returns the calculated `z`, which can also be input to save time.


DavidMStraub:
Great, thanks! I think returning `x, y, z` is indeed a good idea as it decreases the risk of mistakes when saving and loading back plots and changing plot ranges. So I'll merge it as is.

PR #23, opened 2016-11-03
[bll] Changed definition of P, S for B->ll. Also added a test compari…

ChristophNiehoff:
…ng the code to a different implementation.


DavidMStraub:
Great!

PR #27, opened 2016-12-23
Updating L51 of common.py

jackypheno:
Dear Daivd,

I have updated the L51 of flavio/physics/bdecays/common.py, here 'B0': 'bs' is replaced with 'B0': 'bd'.
I request you to update the central repository.

Thanks,
Jacky


DavidMStraub:
Thanks!

PR #29, opened 2017-04-17
add combination of CMS B->K*mumu 2013 and 2015

peterstangl:
@DavidMStraub should the individual `CMS B->K*mumu 2013` and `CMS B->K*mumu 2015` measurements be removed from flavio? (this would break backward compatibility)


peterstangl:
@DavidMStraub I added the 4.3-6 GeV bin of the 2015 measurement separately to use AFB and FL together with the 2017 measurement of P1 and P5p in the same bin. Is this the correct way to do it?

DavidMStraub:
Thanks for the PR! Yes, I think the idea to put the single bin separately is good. Please remove the old measurements - backwards compatibility is not an issue for measurements (superseded measurements are always removed). Please also add
`url: https://twiki.cern.ch/twiki/bin/view/CMSPublic/PhysicsResultsBPH13010`
to the combined measurement.

PR #31, opened 2017-05-26
modifications to qcdf.py and bvgamma.py to make the q2=0 computation possible.

talismanbrandi:
Modifications to qcdf.py and bvgamma.py to make the q2=0 computation possible for b to s gamma exclusive processes. Computation about 5 times faster than using
q2=0.001 GeV^2. Also added ADeltaGamma for B0 to K^*gamma. Can be
removed if not necessary.


DavidMStraub:
Thanks!
At the moment the tests fail because the interpolated version doesn't go down to 0.
Can you please set `q2_arr = np.arange(0, 9, 0.1)` in line 13 of `_qcd_interpolate_write.py` (to remove the shift) and run this script?

Issue #33, opened 2017-06-07
Problem with observables included in "Pseudo-measurement"


peterstangl:
When performing several fits with different combinations of observables, a wrong Warning occurs when one first defines e.g. a global fit and then fits for individual observables. The reason for this is the pseudo-measurement added to the list of measurements by defining the global fit. This pseudo-measurement includes all the observables and thus also constraints all the observables.
Afterwards, trying to fit an individual observable results in the Warning
```python
("{} of the measurements in the fit '{}' "
"constrain both '{}' and '{}', but only the "
"latter is included among the fit "
"observables. This can lead to inconsistent "
"results as the former is profiled over."
).format(count, self.name, obs1, obs2))
```
Explicitly specifying the measurements with `include_measurement` avoids this problem by not including the pseudo-measurements.

DavidMStraub:
Fixed by #34.

PR #34, opened 2017-06-07
exclude pseudo-measurements in "get_measurements" (Issue #33)

peterstangl:
This fixes the bug reported in Issue #33 by excluding all "Pseudo-measurements" from the output of `get_measurements`.


DavidMStraub:
Good point! Thanks!

PR #36, opened 2017-07-28
Update flha.py

wzeren:
I added the Wilson coefficients related to neutral meson mixings, but could not find the proper normalization factor..


DavidMStraub:
Thanks for the PR!

So according to the [FlavorKit paper](https://arxiv.org/pdf/1405.1434.pdf) & the [flavio docs](https://flav-io.github.io/docs/operators.html), the prefactor should be 1.

However, comparing numerically the SM values in the spectrum file contained in flavio with flavio's own SM calculation, I get a relative factor of roughly -2/3, which is very puzzling.

Can you please do the following: take the current verison of SARAH/FlavorKit, take the SUSY scale to a very high value and compute the Wilson coefficient VLL for Bs mixing and tell me the numerical result you get? Thanks!

DavidMStraub:
@avelino3000 FYI :)

wzeren:
Hi David,

we put SUSY scale to roughly 10 TeV and found the following values for VLL of Bs mixing:

 0.56062139E-14    # coeffBsBs_VLLNP

  0.96760906E-10    # coeffBsBs_VLL

0.96755300E-10    # coeffBsBs_VLLSM
Best,
Simon

DavidMStraub:
Thanks!

Sorry, the factor of 0.6 was because I looked at the low-scale coefficient rather than the one at 160 GeV; so this is fine. But I am still puzzled by the minus sign. 

According to the FlavorKit manual, the Lagrangian is *C O*.
According to flavio conventions, the Hamiltonian is *-C O*.
The operators agree, so the Wilson coefficients should, too.
flavio finds a negative coeffBsBs_VLLSM, FlavorKit a positive one.
According to the [Buras review](http://arxiv.org/abs/hep-ph/9806471) eq. (8.9), the Hamiltonian should be (positive number)\**O*, so in the above conventions *C*<0.

wzeren:
@DavidMStraub   Thanks for the reply. In my opinion, the problem of the factor 0.6 was much more serious than the minus sign, so I am happy that it is now gone.

The sign convention in Buras review also agrees with Cheng&Li eq. (12.91),  that the Hamiltonian has a positive prefactor for VLL operator and the Lagrangian has a negative one. This convention then agrees with that of Flavio.
On the other hand, could you please briefly check whether, for *all* five dim-6 operators, the Wilson coefficients found by Flavio have a relative negative sign difference with respect to those found by FlavorKit? If this is the case, then it should have at least no impact on the calculated absolute value of DeltaM's.
Besides, it would be also useful to compare the expression of DeltaM in terms of the Wilson coefficients between Flavio and FlavorKit. Attached is the file listing the observables (including DeltaM's) using MSSM as the model, generated in SARAH4.9.1 by the command MakeSPheno. You may find the expressions of DeltaM's by searching for the keyword "DeltaM".

[FlavorKit_Observables_MSSM.f90.tar.gz](https://github.com/flav-io/flavio/files/1184889/FlavorKit_Observables_MSSM.f90.tar.gz)

DavidMStraub:
Unfortunately I cannot check the other operators since they vanish in the SM. But I find the minus sign also for B_d and K mixing.

Unfortunately even if the sign affects all WCs, this *would* have an impact on DeltaM, as flavio adds the FlavorKit NP contribution to its own internal SM contribution. So the relative signs matter.

The relevant lines in the FlavorKit code seem to be

```fortran
norm = GF**2*mw2/(8._dp*Pi**2)*Veff2
...
CVLLnew = (O4dVLL(3,i1,3,i1)-O4dVLLSM(3,i1,3,i1))/norm
...
Ftt = S0xt(xt) + CVLLnew/(4._dp*r) 
...
ratioDeltaMBq = Abs(Ftt/S0xt(xt))
```

As far as I can see, the sign of this interference means that O4dVLL(3,i1,3,i1) indeed has the opposite sign than what one would expect if the Lagrangian = +*C O*.

wzeren:
I did a similar test and found the same minus sign difference as you got. In SPheno.spc.SM, I manually changed the value of coeffBsBs_VLLNP (both real and imaginary part) from zero to (-1)*coeffBsBs_VLLSM and expected to obtain DeltaM_s ~ zero in Flavio, but only ended up with almost double the original SM-value of DeltaM_s that is hardcoded in Flavio. On the other hand, if I set it to be (+1)*coeffBsBs_VLLSM, the result of DeltaM_d ~0 (11% of the original SM DeltaM_s hardcoded). For B_d, I tested and obtained a similar result.

Now I did another test. Since the hadronic matrix element of VLL and that of VRR have the same expression, I copied coeffBsBs_VLLSM *(+1) to coeffBsBs_VRRNP. This time DeltaM_s gave the exactly same result as when I copied coeffBsBs_VLLSM*(+1) to coeffBsBs_VLLNP. Therefore, I think, VRR also has this minus sign difference, like VLL does. In such a case, I would assume the other operators also have this minus sign difference. I guess, if we set the prefactors of these operators to be (-1) instead of (+1), the problem is solved.

It would be appreciated if you have a better suggestion for checking the other operators.

DavidMStraub:
OK, very good, this confirms that there is indeed an overall sign difference. I think it is important to sort out this sign with the FlavorKit developers to make sure it is consistent also for other processes. I already sent them an e-mail.

PR #47, opened 2018-07-10
align single and multi thread calculations of SM covariance

peterstangl:


peterstangl:
@DavidMStraub, it's not clear to me why `test_measurements_yaml` fails. How can just adding a new test affect any other test? Running `nosetests` locally on my machine, all tests pass.

DavidMStraub:
What happens is that `make_measurement` creates a new instance of `Measurement` (for the pseudo measurement) and this instance for some reason does not pass a test in `test_classes.py` where the instance is converted to YAML and read back in. Probably it would be best to just skip measurements containing the string `Pseudo-measurement` in `test_classes.py` (where also the pathological case `'Belle B->D*lnu hadronic tag 2017'` is being skipped).

DavidMStraub:
Sorry, forgot to merge this.

Thanks!

PR #48, opened 2018-08-02
fix bugs in FastFit with single observable and measurement

peterstangl:
While trying to define FastFit instances using only single observable and measurements, I encountered two bugs:
- The returned mean value was a 1-dim array with one entry instead of just a number.
- Performing the permutations on the SM covariance matrix after loading it from a dictionary lead to an error since this "matrix" is just a number such that no permutations can be applied.

This PR solves those two issues.

### DIFF
--- a/flavio/statistics/fits.py
+++ b/flavio/statistics/fits.py
@@ -509,7 +509,7 @@ class FastFit(Fit):
             covariances.append(covariance)
         # if there is only a single measuement
         if len(means) == 1:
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Very good. Just added a unit test for this case.

peterstangl:
Great, thanks! Is there anything else to do for this PR?

DavidMStraub:
Yes, merging it ;)

PR #49, opened 2018-08-06
[FastFit] allow to save/load experimental central values and covariance

peterstangl:
When defining many `FastFit` instances with several observables and measurements, it might be useful to not only save the SM covariance matrix, but also the experimental covariance matrix and the corresponding experimental central values. The necessary `FastFit` methods to do this are provided by this PR (mainly based on the methods for saving and loading the SM covariance).


DavidMStraub:
Thanks!

PR #51, opened 2018-08-07
Dump wc fit function

peterstangl:
@DavidMStraub this is a first working implementation. However, some things might be done more elegantly, especially on how the schema is used. I'm also not completely happy with the variable name `defdict_fit_wc_function` for the defining dictionary... So suggestions and changes are welcome!

### DIFF
--- a/flavio/statistics/fits.py
+++ b/flavio/statistics/fits.py
@@ -56,7 +56,10 @@ def wc_function_factory(d):
     ```{'code': "def f(C9, C10):\n  return {'C9_bsmumu': 10 * C9, 'C10_bsmumu': 30 * C10}"```
     """
     if 'code' in d:
-        s = d['code']
+        # change the name of the function to 'f'
+        code_lines = d['code'].splitlines()
+        args = code_lines[0].split('(')[1]
+        s = 'def f({}\n{}'.format(args,'\n'.join(code_lines[1:]))
     elif 'args' not in d:
         raise ValueError("Function dictionary not understood.")
     elif 'return' not in d:
@@ -78,7 +81,7 @@ def wc_function_factory(d):
 
 def getsource(f):
     if f is None:
-        return {'code': ''}
+        return None
     try:
         return {'code': inspect.getsource(f)}
     except OSError:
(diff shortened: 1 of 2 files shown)


DavidMStraub:
A problem with this is that it changes the interface of the `Fit` subclasses without any real benefit (the `defdict...` is only useful in connection with the `load` feature, the user otherwise doesn't need to know about it). The number of arguments is already too high. And I would keep the `fit_wc_function` a function.

What about simply

```python
 def load(cls, f):
    kwargs = cls._input_schema(flavio.io.yaml.load_include(f))
    fit = cls(**kwargs)
    if 'fit_wc_function' in kwargs:
        fit._fit_wc_function_string = kwargs['fit_wc_function']
    return fit
```

DavidMStraub:
Correction: one needs to call the `wc_function_factory` in `load` as well (after saving the string to the private attribute).

peterstangl:
Yes, you are completely right. It is a much cleaner solution to put everything in the `load` and `dump` methods and to not change `__init__` at all.

DavidMStraub:
This can be merged now, right?

I still don't quite understand why the `'code'` isn't just taken as-is. The splitting and joining seems a bit hacky. Why should the user be forced to provide both a `def` and `'args'`?

peterstangl:
I don't understand why you write that the user is forced to provide `def` and `'args'`. The user just provides a string with the definition of the function.
If the `code` is taken as-is, the function name of the `fit_wc_function` is defined in the YAML file and this name is arbitrary. After executing the code, this results in defining a function with an unknown name (as long as the name is not retrieved somehow). This means that the `return f` later in the code does not work in general if the function's name is not `f`. The splitting and joining changes the input string such that the name of the function is forced to be `f`.
I admit that this is a bit hacky. Do you know a better solution? E.g. do you know how the function name can be retrieved after defining it using `exec`?

peterstangl:
If one keeps the renaming the function, this could perhaps also be done more elegantly using regex instead of splitting and joining.

DavidMStraub:
Sorry, you're right.

OK if it's just about the renaming I think it's better to force the user to name the function `f`.

PR #52, opened 2018-08-08
use unique function name for fit_wc_function

peterstangl:
With the current implementation, a user-defined function called `f` is just deleted and replaced without any warning. I think it might be useful to use unique function names instead. They are provided by this PR using the `uuid` module.
The following is done:
- The function name in the executable string is called `f_UUID` where `UUID` is a unique hexadecimal identifier.
- After defining the function using `exec(s, globals())`, the local function `f` is defined from the one in the global namespace and the corresponding entry in `globals()` is deleted.
- The local function `f` is returned.

This guarantees that the newly defined function does not overwrite any function in the global namespace and that the global namespace is not populated by many dynamically generated functions.

In addition to its main purpose, this PR also strips whitespaces (including newlines) from the beginning and end of `d['code']`, such that no error is generated if a code string contain e.g. newlines at the beginning.

### DIFF
--- a/flavio/statistics/fits.py
+++ b/flavio/statistics/fits.py
@@ -17,6 +17,7 @@ import pickle
 from functools import partial
 import yaml
 import voluptuous as vol
+import uuid
 
 
 def ensurelist(v):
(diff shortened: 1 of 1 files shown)


peterstangl:
I'm actually not sure if it's really necessary to use a unique name for every imported function, or if it would be enough to just use something less common than `f`.

DavidMStraub:
That's a good point. But I'm still unhappy with the function renaming because it makes assumptions about the form of the source code. What about something as simple as
```python
lambda C9: {'C9_bsmumu': C9}
```
or somebody is crazy enough to use a decorator, etc...

Eventually I think your original suggestion to use dill might be the best option after all because it's clean.
Then one could use a dictionary or source code for the input, but the output would fall back to dill if the is no cached dictionary (or source string).

DavidMStraub:
This is now implemented in #50 so that I think this can be closed.

peterstangl:
@DavidMStraub Yes, I think using dill is a very good solution.

However, there is still the problem that any user-defined function in the global namespace that is called `f` is just overwritten by `wc_function_factory`. But there is quite simple solution to this problem that would also make the renaming completely unnecessary and would allow to load arbitrarily named functions: one can just provide an empty dictionary as a private namespace to `exec`. After calling `exec`, this dictionary only contains two keys: `'__builtins__'` and the second one with the name of the function. This also works for something like
```
myfunc = lambda C9: {'C9_bsmumu': C9}
```
which would lead to a dictionary containing the keys `'__builtins__'` and `'myfunc'` and the value associated to the `'myfunc'` key would just be the function that can be returned. I think this has a huge advantage over using the `globals()` namespace in `exec`.

DavidMStraub:
Great idea! I implemented that. To guard against the (somewhat contrived) case that someone needs an `import` statement before the function definition, I used an `OrderedDict` instead and take the last key as the function.

PR #56, opened 2018-08-14
Update measurements.yml .

jackypheno:


DavidMStraub:
Thanks!

I double-checked it just to be sure. The only case where I'm not sure is `tau->muee`. PDG has &mu;<sup>-</sup>e<sup>+</sup>e<sup>-</sup> and &mu;<sup>+</sup>e<sup>-</sup>e<sup>-</sup>; if I'm not mistaken you  used the number for the latter, but the &tau;&rarr;&mu; process would be the former. Can you please also check again just to be sure?

jackypheno:
You are right Daivd ! I have corrected the value now. But currently in flavio the notation 'BR(tau->muee)'  for the branching ratio  of  tau-  -> mu- e+e- does not specify the lepton charges, I am wondering if at some point  one would like to add also the branching ratio of tau- -> mu+e- e-, we need to some how specify the charges. 

Thanks.

DavidMStraub:
Yes, you're right, it's kind of ambiguous. When in doubt, people can look at the `description` or `tex` attributes, that are also used to generate the [HTML documentation](https://flav-io.github.io/docs/obs/tauleptondecays-lfvdecays.html).

The other decay would probe the operators in the sector we called `etauemu` in WCxf. Not sure if there is a reasonable model generating this, but from on EFT point of view there is no reason not to add it in principle.

PR #57, opened 2018-08-16
add observables and measurement for combined (l1+ l2-, l2+ l1-) LFV BR(B->Ml1l2,l2l1)

peterstangl:
@DavidMStraub, this PR defines the `Observable` instance for the combined emu+mue LFV BR(B->pi(emu+mue)) as used in https://arxiv.org/pdf/hep-ex/0703018.pdf.
To this end, I have put everything used to define the LFV M->Pll `Observable` instances into the function `_define_obs_B_Mll` and removed the currently unnecessary loop `for br in ['BR',]:`. I hope this fine.

Please double-check that my assumption
BR(B->pi(emu+mue)) = BR(B+->pi+ emu) + BR(B+->pi+ mue)
is correct.

Note that like in the last two commits in master, the test introduced in 4056db5df4387a3aea32adbe833a369714ca2604 is failing (`KeyError: 'test_obs 1'`).

### DIFF
--- a/flavio/physics/bdecays/bpll.py
+++ b/flavio/physics/bdecays/bpll.py
@@ -236,7 +236,8 @@ _hadr_lfv = {
 _tex_lfv = {'emu': r'e^+\mu^-', 'mue': r'\mu^+e^-',
     'taue': r'\tau^+e^-', 'etau': r'e^+\tau^-',
     'taumu': r'\tau^+\mu^-', 'mutau': r'\mu^+\tau^-',
-    'emu,mue': r'e^\pm\mu^\mp'}
+    'emu,mue': r'e^\pm\mu^\mp', 'etau,taue': r'e^\pm\tau^\mp',
+    'mutau,taumu': r'\mu^\pm\tau^\mp'}
 
 for l in ['e', 'mu', 'tau']:
     for M in _hadr.keys():
@@ -310,14 +311,11 @@ def _define_obs_B_Mll(M, ll):
     _obs.add_taxonomy(_process_taxonomy)
     return _obs_name
 
-for ll in [('e','mu'), ('mu','e'), ('e','tau'), ('tau','e'), ('mu','tau'), ('tau','mu')]:
-    for M in _hadr_lfv:
+for M in _hadr_lfv:
+    for ll in [('e','mu'), ('mu','e'), ('e','tau'), ('tau','e'), ('mu','tau'), ('tau','mu')]:
         _obs_name = _define_obs_B_Mll(M, ll)
         Prediction(_obs_name, bpll_dbrdq2_tot_func(_hadr_lfv[M]['B'], _hadr_lfv[M]['P'], ll[0], ll[1]))
-
-# Combined emu+mue lepton flavour violating decays
-_obs_name = _define_obs_B_Mll('B+->pi', ('emu,mue',))
-Prediction(_obs_name, bpll_dbrdq2_tot_lfv_comb_func('B+', 'pi+', 'e', 'mu'))
-
-_obs_name = _define_obs_B_Mll('B0->pi', ('emu,mue',))
-Prediction(_obs_name, bpll_dbrdq2_tot_lfv_comb_func('B0', 'pi0', 'e', 'mu'))
+    for ll in [('e','mu'), ('e','tau'), ('mu','tau')]:
+        # Combined l1+ l2- + l2+ l1- lepton flavour violating decays
+        _obs_name = _define_obs_B_Mll(M, ('{0}{1},{1}{0}'.format(*ll),))
+        Prediction(_obs_name, bpll_dbrdq2_tot_lfv_comb_func(_hadr_lfv[M]['B'], _hadr_lfv[M]['P'], ll[0], ll[1]))
--- a/flavio/physics/bdecays/bpll.py
+++ b/flavio/physics/bdecays/bpll.py
@@ -223,7 +223,7 @@ _hadr_lfv = {
 _tex_lfv = {'emu': r'e^+\mu^-', 'mue': r'\mu^+e^-',
     'taue': r'\tau^+e^-', 'etau': r'e^+\tau^-',
     'taumu': r'\tau^+\mu^-', 'mutau': r'\mu^+\tau^-',
-    'emu+mue': r'e^\pm\mu^\mp'}
+    'emu,mue': r'e^\pm\mu^\mp'}
 
 for l in ['e', 'mu', 'tau']:
     for M in _hadr.keys():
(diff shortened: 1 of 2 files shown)


flavio/physics/bdecays/bpll.py
```diff
@@ -281,16 +282,33 @@ def fct(wc_obj, par, q2):
...
-            _obs = Observable(_obs_name)
-            _obs.set_description(r"Total branching ratio of $"+_process_tex+r"$")
-            _obs.tex = r"$\text{BR}(" + _process_tex+r")$"
-            _obs.add_taxonomy(_process_taxonomy)
-            Prediction(_obs_name, bpll_dbrdq2_tot_func(_hadr_lfv[M]['B'], _hadr_lfv[M]['P'], ll[0], ll[1]))
+        _obs_name = _define_obs_B_Mll(M, ll)
+        Prediction(_obs_name, bpll_dbrdq2_tot_func(_hadr_lfv[M]['B'], _hadr_lfv[M]['P'], ll[0], ll[1]))
+
+# Combined emu+mue lepton flavour violating decay B->pi(emu+mue) (hep-ex/0703018)
+# Br(B->pil+l-) = Br(B+->pi+l+l-) = 2 tau_B+/tau_B0 Br(B0->pi0l+l-)
```
DavidMStraub: What is the purpose of this comment?

flavio/physics/bdecays/bpll.py
```diff
@@ -281,16 +282,33 @@ def fct(wc_obj, par, q2):
...
+    mP = par['m_pi+']
+    me = par['m_e']
+    mmu = par['m_mu']
+    q2max = (mB-mP)**2
+    q2min = (me+mmu)**2
+    return (
+        + bpll_dbrdq2_int(q2min, q2max, wc_obj, par, 'B+', 'pi+', 'e', 'mu')
+        + bpll_dbrdq2_int(q2min, q2max, wc_obj, par, 'B+', 'pi+', 'mu', 'e')
+    )*(q2max-q2min)
+_obs_name = _define_obs_B_Mll('B+->pi', ('e(+-)mu(-+)',))
```
DavidMStraub: To be consistent with the corresponding observables in the leptonic decays, the observable should be named `BR(B+->piemu,mue)` (or maybe  `BR(B+->piemu,pimue)` if you think that's more intuitive).

By the way, any reason not to add the neutral decay?

DavidMStraub:
Thanks! I added a few comments.

flavio/physics/bdecays/bpll.py
peterstangl: Should I change it to
```python
# Note that Br(B->pil+l-) is defined in hep-ex/0703018 as
# Br(B->pil+l-) = Br(B+->pi+l+l-) = 2 tau_B+/tau_B0 Br(B0->pi0l+l-)
```
or remove it?

DavidMStraub:
By the way, the failing test is my fault. Will fix it soon, you can ignore it in the meantime.

flavio/physics/bdecays/bpll.py
DavidMStraub: Oh I see. well if it's the isospin average, one shouldn't call it `B+` but simply `B`. But so far for all the LFV modes, both isospin modes are implemented separately. Admittedly this doesn't make too much sense theoretically since there is no good reason for them to be different, but the measurements are often done separately and it could be that some experiment is only sensitive to one of the modes. Thus I suggest to remove the comment and to add the `B0` decay in addition.

peterstangl: OK, I'll add the `B0` decay!

It might then be useful to add this comment to measurement instead of the observable. At least for me it is not very intuitive that, according to hep-ex/0703018, their "isospin averaged branching fraction BR(B→πℓ+ℓ−)" is equal to the `B+` branching fraction.

peterstangl: OK, now I saw that in table II of hep-ex/0703018 the limits for B0 and B+ decays are listed separately. However, now I'm really confused. The table says:
```
BR(B+ -> pi+ e+- mu-+) < 1.7e-7 @ 90% C.L.
BR(B0 -> pi0 e+- mu-+) < 1.4e-7 @ 90% C.L.
BR(B  -> pi  e+- mu-+) < 0.92e-7 @ 90% C.L.
BR(B+ -> pi+ l+ l-) < 1.2e-7 @ 90% C.L.
BR(B0 -> pi0 l+ l-) < 1.2e-7 @ 90% C.L.
BR(B  -> pi  l+ l-) < 0.91e-7 @ 90% C.L.
```
But how does this fit together with their definition of the isospin averaged branching fraction
`BR(B  -> pi  l+ l-) = BR(B+ -> pi+ l+ l-) = 2 tau_B+/tau_B0 BR(B0 -> pi0 l+ l-)`?
The table seems to contradict both of the equal signs in this defintion (givent that according to the PDG `tau_B+/tau_B0 = 1.076 +- 0.004`).

peterstangl:
@DavidMStraub I have added the `B0` decay and implemented both the `B+` and the `B0` decays in such a way that it would now be very easy to also add the combined `l1 l2`+`l2 l1` LFV decays with combinations of leptons in the final state different from `e,mu` as well as decays to kaons instead of pions. Would there be some use of already adding these other decays?

DavidMStraub:
Good question. They are measured separately so it's not really needed, on the other hand since it's trivial to add them I guess it doesn't hurt.

DavidMStraub:
You could also add the measurement from https://arxiv.org/pdf/hep-ex/0703018.pdf while you're at it :)

peterstangl:
@DavidMStraub concerning the measurements, did you read my comment https://github.com/flav-io/flavio/pull/57#discussion_r210670958? Before adding them, it would be nice to understand what's actually going on there. Probably I'm just missing something.

DavidMStraub:
No sorry, I had missed that comment.

The way they write it it's indeed confusing. I would understand it like this:

To quote a combined limit for the isospin modes, they assume there is no isospin asymmetry, which implies ` BR(B+ -> pi+ l+ l-) = 2 tau_B+/tau_B0 BR(B0 -> pi0 l+ l-)` and they call this  `BR(B -> p l+ l-)`. Note that the absence of isospin asymmetry is an assumption for the non-LFV modes which is violated at some level in the SM, but its certainly a valid assumption for the LFV modes.

The table of limits now gives three numbers for each decay mode: the charged decay, the neutral decay, and the limit assuming isospin symmetry to hold. The above relation of course does not apply to these limits. Let's take a simple counterexample: if your detector cannot see pi0s at all, the limit on the B0 mode will be completely gone, and the limit on the BR assuming isospin symmetry will be given directly by the limit on the B+ mode.

By the way, I fear the factor of 2 between the neutral and charged mode is missing in flavio :(

DavidMStraub:
Missing factor of 1/2 added in 0d95eec5a72c2b19a7ce0a1c29e7d4fce0580397

DavidMStraub:
The failing test should now also be fixed in 1dec447aa5748bb14bdc080854760ce7fc0523c0.

peterstangl:
Thanks a lot for the explanation, this makes sense!

I have added the remaining decays and rebased on master.

peterstangl:
I will also add the measurement from https://arxiv.org/pdf/hep-ex/0703018.pdf to this PR

peterstangl:
@DavidMStraub I have also added the two measurements from https://arxiv.org/pdf/hep-ex/0703018.pdf. I have only included the LFV measurements since the non-LFV `BR(B->pill)` is not included as observable at the moment (and the PDG would provide lower limits for these channels than hep-ex/0703018).

DavidMStraub:
Very good!

Yes in fact B->pimumu has been observed and measured by LHCb in the meantime. It is not yet in flavio because the hadronic contributions are really tricky due to |Vud Vus*| ~ |Vtd Vts*| (but they don't enter the LFV decays fortunately).

PR #58, opened 2018-08-17
Adding BR(B0->pitaunu) .

jackypheno:


DavidMStraub:
Thanks!
But they actually have a >2 sigma hint for a signal, so just putting the bound is not entirely consistent. I think the key sentence is

> The signal strength has been chosen such that µ = 1.0 corresponds to B(B0 → π−τ+ντ ) =1.0×10−4. We obtain a best fit of µ = 1.52±0.72

This can be directly treated as a Gaussian constraint.

flavio/data/measurements.yml
```diff
@@ -1894,6 +1894,12 @@ Belle B+->pilnu 2013 total BR:
   values:
     BR(B+->pilnu):  81.2±7.4 e-6
 
+Belle B0->pitaunu 2015 total BR:
+  experiment: Belle
+  inspire: Hamer:2015jsa
+  values:
+  BR(B0->pitaunu): < 2.5e-4 @ 90% CL
```
DavidMStraub: Here there is also a two-space indent missing (that's the reason for the tests failing)

DavidMStraub:
Simply put, the correct code is

```yaml
  values:
    BR(B0->pitaunu): 1.52 ± 0.72 e-4
```

jackypheno:
Right, do we need to take into account the systematics? For example, https://arxiv.org/pdf/1608.05207.pdf takes additional 8% systematic error (see Eq. 1 ).

DavidMStraub:
Oh! This was not so clear to me from the experimental paper, but yes probably this has to be added on top. Please add it in a separate PR as I have already merged this.

PR #60, opened 2018-08-23
Tau->vl

peterstangl:
This PR replaces the old implementation of BR(tau->Vl) that only worked for V=rho0 with a more general one that is also applicable to V=phi.

At the moment, all contributions of operators with vector currents are taken into account as was done in the old implementation. The calculation will be extended in another PR to also consider operators with tensor currents. 

For testing and code review purposes, the old implementation is kept in the code for now and the corresponding predictions are available via the `old_BR(tau->rhomu)` and `old_BR(tau->rhoe)` observables. They should be removed before merging this PR.

Comparing the predictions of `BR(tau->rhomu)` and `old_BR(tau->rhomu)`, I found a difference of <3% for the case where the light lepton mass is set to zero (by replacing `ml` with `0` in line 39, https://github.com/flav-io/flavio/compare/master...peterstangl:tau-Vl?expand=1#diff-4512a505bf1b4ff6faba7c076aec7b12R39). This approximation was used in the old implementation.
In the case where a finite `ml` is kept, I found a difference between  `BR(tau->rhomu)` and `old_BR(tau->rhomu)` of ~7%.

### DIFF
--- a/flavio/physics/taudecays/common.py
+++ b/flavio/physics/taudecays/common.py
@@ -13,10 +13,10 @@ def GammaFvf(M, mv, mf, gL, gR, gTL=0, gtTL=0, gTR=0, gtTR=0):
                 * (2*(mf**2 - M**2)**2 /mv**2 - mf**2 - M**2 - mv**2)
                 - 12 * mf * M * ( (gTL-gtTL).real * (gTR+gtTR).real
                                 + (gTL-gtTL).imag * (gTR+gtTR).imag ) )
-        I = ( 3*M * (mv**2 + mf**2 - M**2)
+        I = ( 3*M * (M**2 - mf**2 - mv**2)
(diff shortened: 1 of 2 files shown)


peterstangl:
Ah yes, good point, I forgot about the dipole contributions...

peterstangl:
After adding a factor `2` in `pre_wc_1` in the old ` br_taurhol` implementation, I find an agreement of the old and new implementations equally good for the dipole operators as for the vector operators (and also for combinations of them).
As soon as the old implementation is not needed anymore for testing purposes, I will remove it. After this, I think the PR should be ready to be merged.

peterstangl:
@DavidMStraub I think this can be merged now.

DavidMStraub:
Thanks @peterstangl & @jackypheno!

PR #62, opened 2018-09-10
B2Dtaunu measurements added

jasonaebischerGIT:


flavio/data/measurements.yml
```diff
@@ -4800,3 +4800,287 @@ CCFR neutrino trident:
   inspire: Mishra:1991bv
   values:
     R_trident: 0.82 ± 0.28
+
+
+
+    Belle B->Dtaunu shape 2015:
```
DavidMStraub: there is a wrong indent here

PR #63, opened 2018-09-16
tau-> pi l implementation. 

jackypheno:


DavidMStraub:
Thanks!

Can you please add a short unit test?

Can't `tau->Ke`, `tau->Kmu` be added easily as well? The only caveat should be factors of sqrt(2) I guess.

jackypheno:
@DavidMStraub OK, I am creating a new pull request.

PR #68, opened 2018-10-16
Bug when loading covariance with several observables for a single observable FastFit

MJKirk:
Bug introduced in pull request #48 -- there is an edge case that if we using FastFit with only a single observable, but are loading a covariance with multiple observables, we get the whole covariance matrix instead of a single element.
Fix by checking the size of the covariance matrix being loaded.

### DIFF
--- a/flavio/statistics/fits.py
+++ b/flavio/statistics/fits.py
@@ -680,7 +680,10 @@ class FastFit(Fit):
         assert len(permutation) == len(self.observables), \
             "Covariance matrix does not contain all necessary entries"
         if len(permutation) == 1:
-            self._sm_covariance = d['covariance']
+            if d['covariance'].shape == ():
+                self._sm_covariance = d['covariance']
+            else:
+                self._sm_covariance = d['covariance'][permutation][:,permutation][0,0]
         else:
             self._sm_covariance = d['covariance'][permutation][:,permutation]
 


MJKirk:
Small thing I noticed while checking my fix, the error message when the loaded covariance doesn't contain all the observables is slightly opaque.
https://github.com/flav-io/flavio/blob/4e0b8623624a5d7d36c8df8a4dde32a19b5d9c96/flavio/statistics/fits.py#L591-L595
The string on line 594 doesn't get printed and the code continues to the next line, and so the only error you get is that permutation is not defined.

DavidMStraub:
Oh, which Exception does it throught then, actually?

MJKirk:
You get UnboundLocalError
The full traceback is
```
Traceback (most recent call last):
  File "flavio_test.py", line 23, in <module>
    f_single_obs.load_sm_covariance("test_covar.p")
  File "/home/mkirk/Documents/Software/flavio/flavio/statistics/fits.py", line 665, in load_sm_covariance
    self.load_sm_covariance_dict(d=data)
  File "/home/mkirk/Documents/Software/flavio/flavio/statistics/fits.py", line 680, in load_sm_covariance_dict
    assert len(permutation) == len(self.observables), \
UnboundLocalError: local variable 'permutation' referenced before assignment
```
So nothing is shown about the covariance matrix missing an observable

DavidMStraub:
Thanks! Nice talk :wink:

DavidMStraub:
Oh, that fix actually doesn't make sense. Shouldn't code during talks.

MJKirk:
https://github.com/MJKirk/flavio/commit/e7cbc0441918547a3a531d332d3bb89a593bc544 is what I did in my fork

DavidMStraub:
Oh exactly, that's what one should do. I can add it.

PR #71, opened 2018-12-11
running for f_perp

peterstangl:
This is a first attempt to implement running for the transverse meson decay constants.


flavio/physics/running/running.py
```diff
@@ -11,6 +11,14 @@
...
 import rundec
 
+# quark mass thresholds
+thresholds = {
+    3: 0.1,
+    4: config['RGE thresholds']['mc'],
+    5: config['RGE thresholds']['mb'],
+    6: config['RGE thresholds']['mt'],
+    7: np.inf,
+    }
```
DavidMStraub: This reduces code duplication, but the problem is that this will be only evaluated once, on module import, and future changes of the `config` dict will have no effect.

flavio/physics/running/running.py
```diff
@@ -302,3 +294,31 @@ def get_wilson(par, c_in, derivative_nf, scale_in, scale_out, nf_out=None):
     sol = rg_evolve_sm(x_in, derivative_nf, scale_in, scale_out, nf_out=nf_out)
     c_out = sol[:-2]
     return c_out.view(np.complex)
+
+def get_f_perp(par, meson, scale):
+    r"""Get the transverse meson decay constant at a given scale.
+    The argument `meson` should be one of `rho0`, `rho+`, `K*0`, `K*+`, `omega`, `phi`
+    """
```
DavidMStraub: We should mention in the doc string that the input is taken from `par` and is assumed to be the value at 1 GeV.

flavio/physics/running/running.py
```diff
@@ -302,3 +294,31 @@ def get_wilson(par, c_in, derivative_nf, scale_in, scale_out, nf_out=None):
...
+    if scale < 0.52:
+        # get_alpha_s yields 0.0 for a scale below 5.2. GeV
+        raise ValueError('RG evolution below 5.2 GeV not implemented.')
```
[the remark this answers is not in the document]
peterstangl: Yes, 0.52 is the value where alpha_s starts becoming 0. I agree with your suggestion. Since `get_alpha_s` is merely a wrapper of `wilson.util.qcd.alpha_s`, would it be even better to check the return value of the latter and raise the error there?

flavio/physics/running/running.py
peterstangl: I addressed this in https://github.com/flav-io/flavio/pull/71/commits/70571a4566bef354484d8363b1725c6cb3ad2c93 by using a function that is called anytime the threshold dictionary is needed. This reduces code duplication and takes into account changes of the `config` dict.

flavio/physics/running/running.py
peterstangl: I addressed this in https://github.com/flav-io/flavio/pull/71/commits/b40cc7f21f8a75b152850651bac4435eb1a1e5fa.

flavio/physics/running/running.py
```diff
@@ -302,3 +294,31 @@ def get_wilson(par, c_in, derivative_nf, scale_in, scale_out, nf_out=None):
...
+            scale_stop = min(thresholds[nf+1], scale)
+            f_perp = _rg_factor_f_perp(par, scale_start, scale_stop, nf)*f_perp
+            if scale_stop == scale:
+                return f_perp
+            scale_start = thresholds[nf+1]
+
+def _rg_factor_f_perp(par, scale_start, scale_stop, nf):
+    alpha_start = get_alpha_s(par, scale_start, nf_out=nf)
+    alpha_stop = get_alpha_s(par, scale_stop, nf_out=nf)
+    return (alpha_stop/alpha_start)**(4/(33-2*nf))
```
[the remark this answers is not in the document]
peterstangl: I addressed this in https://github.com/flav-io/flavio/pull/71/commits/8864fa28beb1f7efb1cf806e7f7b90b5303bd4f7.

flavio/physics/running/running.py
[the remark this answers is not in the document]
DavidMStraub: Yes that would indeed be the most consistent treatment.

DavidMStraub:
Sorry @peterstangl, I forgot to merge this. It is ready, right?

peterstangl:
@DavidMStraub at the moment the implementation still uses the parameter `f_perp` at 1 GeV, while we talked about changing the scale of the `f_perp` parameter to 2 GeV and to use the results from lattice QCD there directly instead of those from the table in https://arxiv.org/pdf/1503.05534.pdf. This is not done yet.

DavidMStraub:
OK, I think that should be it.

PR #72, opened 2019-02-04
[measurements] Update CMS Bs->mumu 2013 and add ATLAS Bs->mumu 2018 measurements

peterstangl:


peterstangl:
> Is it actually correct that the x values start negative?

In principle one could cut off the values at x=0, but I just extended the contours to at least 5 sigma in positive and negative x direction as well as positive y direction, cf. [ATLAS_Bs-mumu.pdf](https://github.com/flav-io/flavio/files/2827951/ATLAS_Bs-mumu.pdf) (axes in the plot are multiplied by 10^-9). This results in the x values starting negative.

DavidMStraub:
OK! The tests pass now as well.

PR #73, opened 2019-02-04
[measurements] Split CMS B->K*mumu 2015 4.3-6 measurem. into BR and ang. obs.

peterstangl:


peterstangl:
@DavidMStraub I think this should be done for the 4.3-6 GeV^2 bin like you did it for the other bins in https://github.com/flav-io/flavio/commit/141ddf372e91b2522c3677bdf9d54cc613f0c53b, right?

DavidMStraub:
Sorry, I forgot why this was split in the first place, can you remind me?

PR #79, opened 2019-03-12
[measurements] add inofficial combination Bs->mumu 2019

peterstangl:


peterstangl:
Note that after this PR, the new flavio version must not be used together with old smelli versions. Otherwise, smelli will combine all Bs->mumu measurements *and* the unofficial combination.

DavidMStraub:
Thanks!

I assume the YAML was generate by the respective method of the probability distribution. In this case it  outputs both the covariance and the correlation. But for the input, it is better to specify only one of them. Can you please just remove the covariance?

DavidMStraub:
NB, I'm in the process of fixing the tests in master

PR #83, opened 2019-03-20
fix bug preventing loading of exp covariances

peterstangl:
The `MeasurementCovariance` class uses the private attribute `_central_cov` to store the central values and covariance matrix. The method `get` tries to return the content of `_central_cov` and recomputes the covariance if this attribute is `None`.
The method `load_dict`, however, had saved the loaded central values and covariance matrix in *another* private attribute `_exp_central_covariance`. So after loading, the attribute `_central_cov` was still `None` and the method `get` recomputed the covariance. This behavior is fixed in the present PR by changing `_exp_central_covariance` in `load_dict` to `_central_cov`.


DavidMStraub:
Oh, shocking. Thanks!

PR #86, opened 2019-03-29
Implementation of neutrinoless mu to e conversion

Broko55:
I implemented the conversion rate for neutrinoless mu to e conversion in Au and Al targets.


DavidMStraub:
That's brilliant, thanks a lot! The implementation looks almost perfect. Just one small thing I noticed: the nucleon charges (`GuS` etc.) should be much better known now than in 2001 (see e.g. appendix 10 of 1902.08191). In fact I have implemented them (in this notation) in a private branch. Let me push this file to the master branch, and then we can rebase your implementation to use these conventions (and numbers) before merging it.

By the way, the numerical parameters (overlap intergrals and `GammaCapture Al`) are currently constants. Can their uncertainties be estimated somehow?

Broko55:
Hi!
In the HFLAG review I only see the axial, tensor and scalar charges. Except for the scalar charges, the related operators do not contribute to the coherent process, that's why I ignored them so far. 

For the capture rate I found this reference https://doi.org/10.1103/PhysRevC.35.2212 giving (gaussian?) uncertainties.
As for the overlap integrals, the only reference where they have been calculated seems to be https://arxiv.org/pdf/hep-ph/0203110.pdf in this scenario and there they do not give any uncertainty estimations. 
Perhaps the thorough thing to do would be implementing their calculation and propagate the nuclear input uncertainties.

At least the uncertainties on GammaCapture I can implement easily and also Ti as a target nucleus.

Furthermore, I noticed that I do not use the charges GxV but rather have them hardcoded, they should be removed from the parameter file.

DavidMStraub:
Indeed the `GxV` can be removed as they are exact integers.

I now pushed the nucleon matrix element function in 545ff8611f306ad31304a8beb91ccd8b7e8488de. The relation to your `GxS` should be:

```python
from flavio.physics.edms.common import proton_charges

pc = proton_charges(par, scale)
GuS = (pc['gS_u+d'] + pc['gS_u-d']) / 2
GdS = (pc['gS_u+d'] - pc['gS_u-d']) / 2
GsS = pc['gS_s']
```

Broko55:
Done!

DavidMStraub:
Great!

The SM uncertainties are currently not too meaningful (since the overlap integrals are constants), but since this is a null test of the SM I think that's not a big problem (until it's actually discovered).

As soon as the next flavio version is released, these can be added to the [smelli](https://github.com/smelli/smelli) global likelihood.

Issue #85, opened 2019-03-26
Flavio with python 3.7: YAMLLoadWarnings


klieret:
Hi David,

I just ran flavio in a python3.7 and got the following YAMLLoadWarnings, thought I'd let you know (this is just a warning, but fixing it is trivial by just replacing ``yaml.load(f)`` with ``yaml.load(f, loader=yaml.FullLoader)``)

```
/srv/conda/lib/python3.7/site-packages/flavio/config.py:5: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  config = yaml.load(f)
/srv/conda/lib/python3.7/site-packages/flavio/measurements.py:14: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  measurements = yaml.load(obj)
/srv/conda/lib/python3.7/site-packages/flavio/parameters.py:15: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  parameters = yaml.load(obj)
/srv/conda/lib/python3.7/site-packages/flavio/parameters.py:29: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  parameters = yaml.load(obj)
/srv/conda/lib/python3.7/site-packages/flavio/parameters.py:42: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  list_ = yaml.load(obj)
```

Or are there other issues with python3.7 (seeing that it's not tested on travis)?

Cheers,
Kilian

DavidMStraub:
Thanks, good to know! PyYAML is a mess...

The reason Python 3.7 is not tested on Travis is that pypmc failed to compile, but probably it's time to deprecate it (unless it is fixed). Indeed assuring 3.7 compatibility is quite important.

klieret:
@MarkusPrim might have been working (or planning to work) on a PR for the Yaml warnings ;)    I might also do it if I find some time in the next weeks (it's super trivial after all).

DavidMStraub:
[...]

PR #87, opened 2019-04-02
Fix yaml warnings

MarkusPrim:
Fixes Issue #85

Get rid of the deprecration warning explained here:
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

The fix is minimal invasive. It explicitly performes what is done implicitly up until now, getting rid of the warning. When you run the tests now, there is still one yaml warning left from the wcxf package. I will create a pull request there to fix the issue.

Cheers,
Markus


DavidMStraub:
Perfect, thanks!

DavidMStraub:
I now realized that `yaml.FullLoader` is actually not defined in prior versions of PyYAML. This makes this fix not backward compatible. While this could be worked around by forcing the new PyYAML version as a dependency, I think this is problematic as many packages now seem to pin the dependency to the *old* PyYAML version to avoid breakages. So I decided to simply replace `load` by `safe_load` everywhere. This should get rid of the warning with PyYAML 5.1, but also work with the old version 3.12. I implemented this in 8e1d05a8ed2e72bb51ab18efff7fa558f20c4ddf.

MarkusPrim:
@DavidMStraub Woops. I did not check for this but relied on the integration build instead. Sorry!

DavidMStraub:
No problem!
I just applied the same fix in wcxf: 413eeb7ede8e8141a37dedd8fc24262b332890bb

PR #91, opened 2019-07-06
Add B(s) -> l l gamma

mreboud:
Implement the B_(s) -> \ell \ell \gamma decays.

### DIFF
--- a/flavio/data/parameters_metadata.yml
+++ b/flavio/data/parameters_metadata.yml
@@ -675,28 +675,28 @@ Gamma12_B0_a:
   description: Coefficient in the approximate equality $\Delta \Gamma_d/M_{12}^d=10^{-4}[c+a\frac{\lambda_u}{\lambda_t}]$
 
 # Parameters for B->gamma form factors
-Bs->gamma KM betav:
+B->gamma KM betav:
   tex: $\beta_v$
   description: beta parameter for Fv form factor in KM parametrization
-Bs->gamma KM deltav:
+B->gamma KM deltav:
   tex: $\Delta_v$
   description: Delta parameter for Fv form factor in KM parametrization
-Bs->gamma KM betaa:
+B->gamma KM betaa:
   tex: $\beta_a$
   description: beta parameter for Fa form factor in KM parametrization
-Bs->gamma KM deltaa:
+B->gamma KM deltaa:
   tex: $\Delta_a$
   description: Delta parameter for Fa form factor in KM parametrization
-Bs->gamma KM betatv:
+B->gamma KM betatv:
   tex: $\beta_{tv}$
   description: beta parameter for Ftv form factor in KM parametrization
-Bs->gamma KM deltatv:
+B->gamma KM deltatv:
   tex: $\Delta_{tv}$
   description: Delta parameter for Ftv form factor in KM parametrization
-Bs->gamma KM betata:
+B->gamma KM betata:
   tex: $\beta_{ta}$
   description: beta parameter for Fta form factor in KM parametrization
-Bs->gamma KM deltata:
+B->gamma KM deltata:
   tex: $\Delta_{ta}$
   description: Delta parameter for Fta form factor in KM parametrization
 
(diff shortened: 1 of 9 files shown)


flavio/data/parameters_metadata.yml
```diff
@@ -674,6 +674,43 @@ Gamma12_B0_a:
...
+Bs->gamma KM betata:
+  tex: $\beta_{ta}$
+  description: beta parameter for Fta form factor in KM parametrization
+Bs->gamma KM deltata:
+  tex: $\Delta_{ta}$
+  description: Delta parameter for Fta form factor in KM parametrization
+
+
+# Parameters for the Bs->phi form factors (1503.05534v3)
+Bs->phi LCSR T10:
```
DavidMStraub: This shouldn't be needed as `Bs->phi` form factors are already implemented in flavio. This parameter should actually be equal to the existing parameter `Bs->phi BSZ a0_T1` (which does not show up in this file as it is created programatically in `bsz.py`, but shows up on the [parameter list](https://flav-io.github.io/docs/parameters.html).

flavio/data/parameters_metadata.yml
```diff
@@ -674,6 +674,43 @@ Gamma12_B0_a:
...
+  tex: $T_1(0)$
+  description: Bs->phi T_1 parameter at q2=0 
+Gamma_phi:
+  tex: $\Gamma_\phi$
```
DavidMStraub: This shouldn't be necessary as there is already a parameter `tau_phi` (the lifetime) that is just the inverse width.

flavio/physics/bdecays/bllgamma.py
```diff
@@ -0,0 +1,168 @@
...
+        _process_tex =  r"B_s\to" +_tex[l]+r"^+"+_tex[l]+r"^-\gamma"
+        _process_taxonomy = r'Process :: $b$ hadron decays :: FCNC decays :: $B\to \ell^+\ell^-\gamma$ :: $'
+
+
+        # binned branching ratio
+        _obs_name = "<dBR/dq2>(Bs->"+l+l+"gamma)"
+        _obs = Observable(name=_obs_name, arguments=['q2min', 'q2max'])
+        _obs.set_description(r"Binned differential branching ratio of $" + _process_tex + r"$")
+        _obs.tex = r"$\langle \frac{d\text{BR}}{dq^2} \rangle(" + _process_tex + r")$"
+        _obs.add_taxonomy(_process_taxonomy)
```
DavidMStraub: Here probably you want to add the `_process_tex` and be careful with dollar signs (there is a stray one at the end of `_process_taxonomy`)

flavio/data/parameters_metadata.yml
mreboud: Exact, I will change this!

flavio/data/parameters_metadata.yml
mreboud: Here again, I will make the change. (I made these naive adds to have a working implementation)

flavio/physics/bdecays/bllgamma.py
mreboud: Did you mean trading ```_process_taxonomy``` for ```_process_taxonomy + _process_tex + r"$"```?

DavidMStraub: Looks good, yes.

mreboud:
One question: Charmonium resonances are not implemented, which means that the user should not use a ```[q2min, q2max]``` range that includes them. Should I add a warning if the range is ill-defined?

DavidMStraub:
Yes, a warning is probably a good idea (I did something similar for the QCDF corrections to B->K*ll).

mreboud:
Apart from a test file, the implementation is complete. I wanted to add BaBar measurement (http://inspirehep.net/record/753579) but what they actually measured is not clear. They don't specify the q^2 range but say that they vetoed the charmonium resonances...

DavidMStraub:
@mreboud, just to be sure I'm not forgetting to merge this: do I understand correctly you still plan to add a unit test before I should merge it? Thanks!

mreboud:
You are right, I have been busy with other things and kind of forget to finish this. I will do it asap.

mreboud:
I added a basic test file and did a few cross checks. I think that the code is ready to be merged unless you have any comments.

DavidMStraub:
Looks great, thanks!

PR #92, opened 2019-09-02
New B->D(*) form factors

DavidMStraub:
An attempt to implement the form factors from https://arxiv.org/abs/1908.09398 as new defaults.

@dvandyk
@MartinJung1978


flavio/physics/bdecays/formfactors/b_v/cln.py
```diff
@@ -66,30 +66,34 @@ def ff(process, q2, par, scale):
...
+    xi3 = par['CLN xi3']
+    xi = hqet.xi(z, rho2, c, xi3, order_z=2)
     h = {}
     h['V'] = xi * (1 + ash * CV1
                    + epsc * (L[2] - L[5])
                    + epsb * (L[1] - L[4]))
     h['A1'] = xi * (1 + ash * CA1
                     + epsc * (L[2] - L[5] * (w - 1)/(w + 1))
                     + epsb * (L[1] - L[4] * (w - 1)/(w + 1))
-                    + epsc**2 * par['B->D* CLN deltac_hA1'])
```
DavidMStraub: @dvandyk, @MartinJung1978 do I have to expand the `(w-1)/(w+1)` in z as well?

dvandyk: @DavidMStraub No. We keep all ```w``` terms without z-expansion if they are of kinematical origin.

DavidMStraub: :+1: for that.

So far the only places where I expand is in \xi (this is the 3 in 321, right?) and in \ell_i + (w-1)\ell'_i (this is the 1, right?) Where is the 2 in 321 then?

dvandyk: >  for that.
> 
> So far the only places where I expand is in \xi (this is the 3 in 321, right?)
correct!

> and in \ell_i + (w-1)\ell'_i (this is the 1, right?)
Correct.

> Where is the 2 in 321 then?
When you expand the subleading IW functions (chi_2, chi_3 and eta), you have to expand to 2nd order in (w- 1), and that to 2nd order in z.

DavidMStraub: Oh, good, still need to add them anyway. Thanks.

DavidMStraub: Good, I believe the 321 is implemented correctly now in 05aa991ce8fd91dd9ca3c4e7d4e7a0167c210f28.

dvandyk:
@DavidMStraub here are some test cases that were verified by Martin:

For B->D: https://github.com/eos/eos/blob/master/eos/form-factors/mesonic-hqet_TEST.cc#L199-L358
For B->D*: https://github.com/eos/eos/blob/master/eos/form-factors/mesonic-hqet_TEST.cc#L1023-L1180

Note that in both cases the model is 3/2/1. We tested with a different set 2/2/1, too.

DavidMStraub:
Experience shows that having numerical values of the FFs, maybe at some intermediate q² value, in the standard basis, would be a powerful cross-check of the implementation and could serve as a unit test. I suppose this is fairly easy to generate with EOS, am I right @dvandyk?

DavidMStraub:
Wait. Did you just answer my question 3 minutes before I asked it?? :open_mouth:

dvandyk:
> Wait. Did you just answer my question 3 minutes before I asked it??

I'm just that good! *drops keyboard*

dvandyk:
Sorry, there is a typo in my second-to-last message. The model for the links is '3/1/1', not '3/2/1'. Not quite sure, how easily you can change the model for test purposes.

DavidMStraub:
Also, do I see correctly I need to add a parameter xi''''?

dvandyk:
> Also, do I see correctly I need to add a parameter xi''''?

Only if you want to support a 4/l/m model. EOS does, but it was not neccessary for the fit to all theory inputs.

DavidMStraub:
OK great thanks to a productive session with @dvandyk, the unit tests work now! The only problem is very strong sensitivity of the unit test to the precise values of alpha_s and m_b, and the fact that the default parameter values for alpha_s, m_b, and lambda_1 are different from the ones used in your fit (which, strictly speaking, makes using the fitted parameters inconsistent). But I think this is mostly due to the extreme values in the unit test and should be OK to neglect in practice. Agreed?

dvandyk:
Excellent!

PR #100, opened 2019-11-22
[classes] add `read_default` method to `ParameterConstraints` class

peterstangl:


DavidMStraub:
Good idea. Comments:

- It might be a bit confusing that e.g. for `Measurement`, `clear_all` is a *class method* that deletes all instances (and not acting on an instance), while for `ParameterConstraints`, it would be an instance method flushing the instance. So I would suggest renaming it.
- I understand the use case for `read_default` much better than for `clear_all`. To reset an instance, it seems much more Pythonic and predictable to me to create a new one, or even to repeat the init, e.g.
```python
flavio.default_parameters.__init__()
```
I find it confusing to add a new method that does nothing else than `__init__`. By the way: the constructor also resets the constraints dict, which your instance method doesn't.
- I suggest to reset the instance within the `read_default` method
- Don't forget doc strings :)

PR #102, opened 2019-12-15
Update measurements.yml

afalkows017:
update of LEP-1 measurements due to luminosity update, see [arXiv:1912.02067]


peterstangl:
@afalkows017 thanks a lot!

From a brief look at arXiv:1912.02067, I understand that this paper updates LEP measurements based on a more precise determination of the integrated luminosity. Apart from the updated values, there are also several other LEP measurements in `measurements.yml`: https://github.com/flav-io/flavio/blob/63e602cf38fce48745f8acd225efe4e7bbc55156/flavio/data/measurements.yml#L5837-L5889
So I wonder if it is fully consistent to update some of the measurements while keeping others unchanged. I assume that it is actually consistent since the update mostly concerns the total width of the Z and its hadronic cross section. The other observables are ratios that should be essentially unaffected.
Do you agree?

afalkows017:
Hi Peter,
most LEP-1 measurements are ratios so they are not affected by the
luminosity change.
The one that is affected is sigma_had because it's an absolute cross
section measurement and not a ratio.
There's also a tiny shift of the central value of Gamma_Z, i guess due to
correlations.
The remaining measurements are not  affected.
1912.02067 give the full list of measurements and the new correlation
matrix in Table B13
I updated everything that changed.

best
Adam

peterstangl:
@afalkows017 great, thanks for the confirmation!
I have updated the inspire id and will merge this PR.

Issue #112, opened 2020-07-14
scikit-hep


DavidMStraub:
@peterstangl, in feedback from a talk at PyHEP, scikit-hep's https://github.com/scikit-hep/decaylanguage and https://github.com/scikit-hep/particle have been mentioned as projects that flavio could profit from. Would be interesting, e.g. to get rid of [this](https://github.com/flav-io/flavio/blob/master/flavio/parameters.py#L127).

eduardo-rodrigues:
Thanks for this follow-up. I'm happy to help. BTW, I'm working on improving Particle so you should let me know what you need in case there are things not there yet ... (Probably the biggest change to come is that I'm adding "all" particles for which the PDG does not give info in their downloadable file, so BSM, etc. A lot is there in a way, but not via the trivial and common usage.)

eduardo-rodrigues:
BTW, great presentation with many thoughts straight to the point. I think you hit the right mode for such a keynote presentation. Thanks again for participating!

peterstangl:
@DavidMStraub thanks for the suggestion! It looks like the Particle package would really be a perfect replacement for the functions currently used for reading particle masses and widths from the PDG.
@eduardo-rodrigues thank you for offering your help! One thing I was wondering is whether Particle always provides the most recent PDG data or if it is also possible to get older PDG data, e.g. from a given year. This would be useful if we want each version of flavio to use a fixed set of input parameters (in order to facilitate reproducibility).

eduardo-rodrigues:
Hello @peterstangl. By default `Particle` provides the most recent PDG data file, which is presently the one from 2019, see `https://github.com/scikit-hep/particle/tree/master/src/particle/data`. I now checked and the 2020 PDG file is actually available, see http://pdg.lbl.gov/2020/mcdata/mass_width_2020.mcd, hence I will make it soon available in `Particle`.

This being said, the package internally uses an extended file prepared by us, which contains more info than what the PDG .mcd file contains. That's the particleYYYY.csv file under the same location. The standard user never has to care about or interact with the CSV file(s).

I did a clean-up recently and only left in the package the data corresponding to 2018. We could provide older files for convenience ... Just let us know.

In any case, it is easy to do a custom loading. For a standard usage do for example

```
In [1]: import particle
In [2]: particle.__version__
Out[2]: '0.10.0'
In [3]: from particle import Particle
In [4]: Particle.table_loaded()
Out[4]: True
In [5]: Particle.table_names()
[...]

PR #114, opened 2020-07-31
use `particle` package to get PDG data 

peterstangl:
This PR replaces the functions for reading PDG data files by an interface to the python package `particle` as suggested and discussed in isse #112.

The main changes are:
- Particles are now identified by their `flavio_name`, which is the name used e.g. in the mass parameters, and by their PDG IDs. Translating the former into the latter is done using the `pdg_particles` dictionary, replacing the old lists `pdg_include` and `pdg_translate`.
- Masses and widths are now obtained from `particle`. PDG data files and the old function `_read_pdg_masswidth` are removed.
- The latex names are now  obtained from `particle`, but simplified using the new `_pdg_tex_simplify` function. The result is equivalent to the one of the old `_pdg_particle_string_to_tex` function, which is removed.
- New tests are added that check the masses and latex names of all particles currently contained in `pdg_particles`.


flavio/parameters.py
```diff
@@ -91,168 +92,127 @@ def write_file(filename, constraints):
...
+    r"(?:\^\{(.*?)\})*" # ^{...}
+)
+
+def _pdg_tex_simplify(string):
+    m = _pdg_tex_regex.match(string)
+    name = m.group(1)
+    sub = m.group(2)
+    sup = (m.group(3) or '') + (m.group(5) or '')
+    par = m.group(4)
+    if sub or name in 'W Z H e \\mu \\tau':
```
DavidMStraub, line 155: Shouldn't this be `name in ["W", "Z", ...]`? I think `"\\mu" in "\\mu"` is `False` since it iterates character by characters.

flavio/parameters.py
```diff
@@ -91,168 +92,127 @@ def write_file(filename, constraints):
...
+        # remove superscripts +-0 and keep only *
+        sup = '*' if '*' in sup else ''
+    if not sub and par and par.upper().isupper() and par not in '1S':
```
DavidMStraub, line 158: `not in "1S"` matches everything except `1` and `S`

DavidMStraub:
Great! Possibly all the particle-specific stuff could be collected in a class (e.g. `class FlavioParticle(Particle)`), which would allow to make the `read_pdg` much more concise. Just a thought.

flavio/parameters.py
peterstangl: Good point, I think this should maybe better be `... and name != 'J/\\psi'`.
The idea was to add a subscript width what is inside the parentheses of the latex name in case this is more than just a number (usually corresponding to the mass) and if there is no other subscript. So e.g. `'D^{*}(2007)^{0}'` becomes`'D^{*0}'` since the parentheses contain only a number, but `'\\psi(2S)'` becomes `'\\psi_{2S}'` since the parentheses do not contain only a number. (Checking the content of the parentheses is currently done using `par.upper().isupper()` but should probably better be done by `not par.isdigit()`)
The only exception from this rule is `'J/\\psi(1S)'`, which should become `'J/\\psi'`. So I guess it's clearer to apply this exception only to `'J/\\psi'` and not to anything containing `1S`.

flavio/parameters.py
peterstangl: `"\\mu" in "\\mu"` is `True` since `in` tests for substring containment (cf. https://docs.python.org/3/reference/expressions.html#membership-test-details).
But I agree that it's maybe not a good idea to just test for substring containment since e.g. also names `a`, `u`, and `au` would be `in '\\tau'`.

DavidMStraub: Oh, you're right, sorry! But yeah, list might still be better.

peterstangl:
> Great! Possibly all the particle-specific stuff could be collected in a class (e.g. `class FlavioParticle(Particle)`), which would allow to make the `read_pdg` much more concise. Just a thought.

This is a very good idea and I have now moved everything to the new `FlavioParticle` class in commit da46a40.
Since it is a class, should it be moved to `flavio.classes`?
I'm also not sure what to do with the `pdg_particles` dictionary. In principle this is more like a config setting that one could move to a yaml file. Do you have a good idea in which file this could fit?

DavidMStraub:
Nice! 

> Since it is a class, should it be moved to flavio.classes?

I think it's fine... `classes` is anyway a bit of a silly name for a module. There are also classes in other modules like `probability`.

> I'm also not sure what to do with the pdg_particles dictionary. In principle this is more like a config setting that one could move to a yaml file. Do you have a good idea in which file this could fit?

It's a constant and it's only needed by the class, so I think it naturally fits as a class attribute:

```python
class FlavioParticle(Particle):
    """..."""

    PDG_PARTICLES = {...}
```

which can be accessed as `cls.PDG_PARTICLES` in the class methods and `self.PDG_PARTICLES` in the instance methods.

peterstangl:
> It's a constant and it's only needed by the class, so I think it naturally fits as a class attribute:

Yes, I agree and have implemented this in 0582522.
In 2b8e496, I now also added the class method `flavio_all`, which returns a set of all instances used in flavio. This simplifies looping over the particles, as e.g. done in the `read_pdg` function.

This now looks quite complete and I would merge this PR soon.

PR #119, opened 2020-10-08
Update of experimental measurements

olcyr:
Inclusion of ATLAS measurements of W->tau nu (https://arxiv.org/abs/2007.14040) and new limits on Z LFV (https://arxiv.org/abs/2010.02566).

### DIFF
--- a/flavio/data/measurements.yml
+++ b/flavio/data/measurements.yml
@@ -2873,6 +2873,29 @@ LHCb Bs->phimumu 2015 15.0-19.0 GeV^2:
                 [1.00, 0.03],
                 [1.00]]
 
+LHCb B->K*ee 2020:
+  experiment: LHCb
+  inspire: Aaij:2020umj
+  values:
+    - name: <FL>(B0->K*ee)
+      q2min: 0.000784
+      q2max: 0.257
+      value: 0.044 ± 0.026 ± 0.014
+    - name: <P2>(B0->K*ee)
+      q2min: 0.000784
+      q2max: 0.257
+      value: -0.03 ± 0.04 ± 0.01 # P2 = 1/2 * ATRe
+    - name: <P1>(B0->K*ee) 
+      q2min: 0.000784
+      q2max: 0.257
+      value: 0.11 ± 0.10 ± 0.02 # P1 = AT2
+    - name: <ATIm>(B0->K*ee)
+      q2min: 0.000784
+      q2max: 0.257
+      value: 0.02 ± 0.10 ± 0.01
+  correlation:
+    [[1, -0.02, -0.01, 0.02], [-0.02, 1, 0.05, 0.02], [-0.01, 0.05, 1, 0.1], [0.02, 0.02, 0.10, 1]]
+
 LHCb B->K*ee 2013:
   experiment: LHCb
   inspire: Aaij:2015dea
@@ -6011,7 +6034,7 @@ DELPHI Z LFV:
       counts_total: 0
       counts_background: 0
 
-ATLAS Z LFV 2020:
+ATLAS Z->taul 2020:
   experiment: ATLAS
   inspire: Aad:2020gkd
   values:
--- a/flavio/data/measurements.yml
+++ b/flavio/data/measurements.yml
@@ -5539,6 +5539,12 @@ LHCb B->ll' 2017:
     BR(Bs->emu,mue): < 6.3e-9 @ 95% CL
     BR(B0->emu,mue): < 1.3e-9 @ 95% CL
 
+LHCb B->Ktaumu 2020:
+  experiment: LHCb
+  inspire: Aaij:2020mqb
+  values:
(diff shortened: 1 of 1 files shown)


peterstangl:
Great, thanks!

If the ATLAS Z LFV 2020 measurement also contains the data from the ATLAS Z LFV 2014 measurement, then I think we can remove the 2014 one.

olcyr:
> Great, thanks!
> 
> If the ATLAS Z LFV 2020 measurement also contains the data from the ATLAS Z LFV 2014 measurement, then I think we can remove the 2014 one.

Thanks, Peter! ATLAS Z LFV 2014 was a search for for Z-> mu e. The new paper is for Z -> mu tau and Z -> e tau, so I think we should leave as it is.

peterstangl:
> Thanks, Peter! ATLAS Z LFV 2014 was a search for for Z-> mu e. The new paper is for Z -> mu tau and Z -> e tau, so I think we should leave as it is.

Ah, yes of course you are right. I was only looking at the name...

DavidMStraub:
In fact `ATLAS Z->taul 2020` might be more fitting, in case they also release a new `emu` measurement.

flavio/data/measurements.yml
```diff
@@ -5539,6 +5539,12 @@ LHCb B->ll' 2017:
     BR(Bs->emu,mue): < 6.3e-9 @ 95% CL
     BR(B0->emu,mue): < 1.3e-9 @ 95% CL
 
+LHCb B->Ktaumu 2020:
+  experiment: LHCb
+  inspire: Aaij:2020mqb
+  values:
+    BR(B+->Ktaumu): < 3.9e-5 @ 90% CL # NB this is B- -> K- tau+ mu-
```
peterstangl, line 5546: @olcyr @DavidMStraub I'm a bit confused by the definitions here. So in the flavio notation there is no `B-` but only a `B+` and so the name `B+->Ktaumu` corresponds to `B- -> K- tau+ mu-`.

But what happens if one has `B+ -> K+`? On the level of the operator, `B- -> K- tau+ mu-` corresponds to something like `(sbar ga_nu b)(mubar ga^nu tau)`, right? And then for `B+ -> K+ tau- mu+` I would think that one has just the Hermitian conjugate of that operator, namely `(bbar ga_nu s)(taubar ga^nu mu)`, right?
This would mean that measurements of both `B- -> K- tau+ mu-` and `B+ -> K+ tau- mu+` constrain the same quark-level operator (up to Hermitian conjugation). Is that correct?
So if I exchange `B-` with `B+`, I would think that I also have to exchange `tau+ mu-` with `tau- mu+` if I want to constrain the same operator.

Now what makes me confused is that LHCb has measured `B+ -> K+ mu- tau+` and if my reasoning above is correct, this should correspond to `B- -> K- mu+ tau-`, which in flavio is `B+->Kmutau`. However, here and also in the 2012 BaBar measurement shown below, the measurement of `B+ -> K+ mu- tau+` is associated with `B+->Ktaumu`.

At what point do I get it wrong?

https://github.com/flav-io/flavio/blob/ea735bd8febbb961d249eddf338a4960c1fbee69/flavio/data/measurements.yml#L5497-L5508

DavidMStraub: > This would mean that measurements of both B- -> K- tau+ mu- and B+ -> K+ tau- mu+ constrain the same quark-level operator (up to Hermitian conjugation). Is that correct?

Yes. They are CP conjugates. And since there are no strong phases involved, we can assume the rates to be equal (this is a bit less trivial than it sounds IMO).

>  However, here and also in the 2012 BaBar measurement shown below, the measurement of B+ -> K+ mu- tau+ is associated with B+->Ktaumu.
>
> At what point do I get it wrong?

I believe you are right and the BaBar assignment is indeed wrong. Can you please check if it is consistently wrong also for the other decays in this measurement, and whether it is also wrong for other measurements?

BTW, I am very confident the charges shown in the [docs](https://flav-io.github.io/docs/obs/bhadrondecays-fcncdecays.html) are actually correct so that can be taken as a reference.

peterstangl:
sorry, I didn't want to close this PR, but just to rebase the branch on the current flavio master

flavio/data/measurements.yml
peterstangl: It seems that in all of the following Babar measurements, the leptons should be exchanged https://github.com/flav-io/flavio/blob/ea735bd8febbb961d249eddf338a4960c1fbee69/flavio/data/measurements.yml#L5497-L5508

And also in the measurement newly added by @olcyr, it should be `mutau`.

I have checked other LFV measurements and there the leptons seemed to be assigned correctly. The only thing I noticed is that in the following LHCb measurement, the `B+->Kemu` bound should be `6.4e-9 @ 90% CL` (I guess this measurement is https://cds.cern.ch/record/2688536 and the inspire ID is Aaij:2019nmj) https://github.com/flav-io/flavio/blob/ea735bd8febbb961d249eddf338a4960c1fbee69/flavio/data/measurements.yml#L5549-L5553

PR #122, opened 2020-10-14
Some LEP measurement updates 

MJKirk:
I've added some LEP Z and W leptonic decay measurements which were missing, and which we decided to include to make flavio more complete even at the expense of having a consistent set of data.

Also, one measurement didn't have an inspire reference (the only one if my regexp was correct), so I added that.


peterstangl:
Thanks @MJKirk! I think this can be merged.

Issue #138, opened 2021-01-21
Errors in implementation of K0->ll'


hoodyn:
Hi, 

I think I've found two problems in the implementation of K0->ll' in flavio/physics/kdecays/kll.py.

### Lepton flavour violating K-short decay results in errors
`K=KS` together with `l1 != l2` not a valid option in the definition of function `amplitudes_eff`:
```
   ...
    if K == 'KS' and l1 == l2:
        Peff = P.imag
        Seff = S.real + SLD
    if K == 'KL':
        Peff = P.real + PLD
        Seff = S.imag
    return Peff, Seff
```
Correspondingly, asking flavio for a prediction on `BR(KS->emu,mue)` leads to 
`UnboundLocalError: local variable 'Peff' referenced before assignment`.

This would have been trivial to fix, be there not the second problem. 

### Lepton flavour violating K-long decay calculated incorrectly
This is physIcs-related issue: the formulas for using imaginary of real parts of the Wilson coefficients (or, of the `P,S` parts of the `amplitudes` in the current flavio implementation) hold only for the lepton flavour conserving case. 

Generally, KL and KS are superpositions of K0 and antiK0: K_L,S = (sbar d +- dbar s) / sqrt(2). 
Thus, naively, the amplitudes consist of the combinations `(CX_sdl1l2 +- CX_dsl1l2)/sqrt(2)` with X=9,9p,10,10p,S,SP,P,Pp.
However, in flavio, the `_dsl1l2` Wilson coefficients are not to be specified as they are obtained from complex-conjugated `_sdl2l1` ones. In particular: 
![image](https://user-images.githubusercontent.com/77328256/105366951-33939f00-5c00-11eb-9022-cd777b39f038.png)
(The extra minus sign for `CS-CSp` is spit by the CP transformation of the corresponding operator - I would appreciate if someone could check the above formulas independently. The  `CX + CXp` combinations are irelevant for pseudoscalar meson decays.)
If `l1==l2`, one gets terms like `( (CX_sdll-CXp_sdll) +- (CX_sdll-CXp_sdll).conj() ) / sqrt(2)` which simplify to real or i*imaginary parts indeed.
However, the LFV cases lead to expressions like `( (CX_sdemu-CXp_sdemu) +- (CX_sdmue-CXp_sdmue).conj() ) / sqrt(2)` which obviously cannot be simplified any further.

Regards, Matej Hudec

hoodyn:
[...]

PR #140, opened 2021-01-29
K to ll' revision (fix to #138)

hoodyn:
In the discussion in #138, we agreed that there are some bugs in Kll.py. I tried to repair them. Moreover, I commited some other small changes aiming to improve the readability of the code. 

1. The KS->e mu is no longer an invalid option in the function `amplitudes_eff`. 

2. Amplitudes for LFV decays (i.e. `l1!=l2`) were not calculated correctly. The fix needed some nontrivial extension of the code as well as some physics considerations, which I will summarize now:
-  the original function function `amplitudes` has been renamed to `amplitudes_weak_eigst` as it actually does not yield the relevant amplitudes for KL -> l1+l2- nor KS->l1+l2-, but for the decay of weak eigenstate K0bar -> l1+ l2-. 

```
   S_K0bar, P_K0bar = amplitudes_weak_eigst(par, wc, l1, l2)
```
Also the amplitude for K0->l1+l2- can be obtained from this function as
```
    S_aux, P_aux = amplitudes_weak_eigst(par, wc, l2, l1)
    S_K0 = -S_aux.conjugate()
    P_K0 = P_aux.conjugate()
```
I can attach some notes about why it is like this.
- An entirely new function `amplitudes(par, wc, K, l1, l2)` has been implemented. It simply makes proper linear combinations of amplitudes for K0 and K0bar which correspond to the approximate mass eigenstates KL, KS (in fact, CP eigenstates are used instead). In the special case `l1 == l2`, this can be simplified to the expressions with imaginary or real parts of S,P, which had been there originally, adopted from ArXiv:1711.11030 (I guess). I implemented this special case separately inside the `amplitudes`, for the reason of both effectivity and better possibility to check that the new formulae are consistent with the old ones. If the maintainers find this separate treatment  redundant, it can be removed, of course. The new  `amplitudes` function now takes `K` as a mandatory argument (to be `'KL'` or `'KS`'). 
- There is no longer a need to take real or imaginary parts of the `amplitudes` within the `amplitudes_eff` function, which now serves purely to sum the short-distance `amplitudes` with `amplitudes_LD` in the LF conserving cases. It also seems more consistent now - `amplitudes_eff` has one and only one purpose.

Furthermore, I made some additional small commits:
3. The function `amplitudes_LD(par, K, l)` was originally giving long-distance contributions for both KL and KS, although it takes `K` as an argument. Eventually, everything had been calculated correctly but the implementation seemed little awkward to me. So I've changed it in order to make things clearer.
4. I dared to change the normalization of the new short-distance `amplitudes` and `amplitudes_LD` by `sqrt(2)` and reduced the prefactor of `br_kll` by 2 accordingly. The purpose is solely to make `br_kll` more consistent with `br_inst` in ./physics/kdecays/kll.py or with arXiv:1602.00881 from which the formulae have been adopted; also, the normalization of `amplitudes` as linear combinations of `amplitudes_weak_eigst` is the standard one, following from KL = (K0+K0bar)/sqrt(2),KS = (K0-K0bar)/sqrt(2).

I also think I may have found another bug, which we haven't discussed in #138. 
5. Originally, `amplitudes_eff` for the case `K=='KS'` was was summing the effective scalar amplitude as `Seff = S.real + SLD`. However,  ArXiv:1711.11030 (which is the cited source in the code) claims that these two terms should subtract from each other (overall phase is irrelevant). `SLD` is implemented in flavio correctly according to Eq. (2.15) in 1711.11030 including the sign. Also, both flavio and 1711.11030 use the same convention for WCs: _L_eff=-H_eff=C_A O_A_. Thus, I think that originally the interference was inappropriately implemented. In my proposal it is corrected by the fact that the short-distance amplitude `S` is obtained as `-S_K0bar.real * sqrt(2)`. (The minus sign is the important thing there; the sqrt(2) relates to what I discuss above in point 4.). Note that both flavio and 1711.11030 use the same convention for WCs: _L_eff=-H_eff=C_A O_A_.
The similar issue for the case of KL (`Peff = P + PLD`) was (and still is) probably correctly handled by the extra minus sign in the definition of PLD:
```
        ALgaga = 2.02e-4 * (par['chi_disp(KL->gammagamma)'] - 5.21j)
        SLD = 0
        PLD = -pre * ALgaga
``` 
I didn't dig to details there (like what's the default sign of the parameter enteging ALgaga in flavio when in fact it is not known).

### DIFF
--- a/flavio/physics/kdecays/kll.py
+++ b/flavio/physics/kdecays/kll.py
@@ -9,15 +9,13 @@ from flavio.physics.common import add_dict
 from math import pi, sqrt
 
 
-def amplitudes(par, wc, l1, l2):
-    r"""Amplitudes P and S entering the $K\to\ell_1^+\ell_2^-$ observables.
+def amplitudes_weak_eigst(par, wc, l1, l2):
+    r"""Amplitudes P and S for the decay $\bar K^0\to\ell_1^+\ell_2^-$.
 
     Parameters
     ----------
-
     - `par`: parameter dictionary
     - `wc`: Wilson coefficient dictionary
-    - `K`: should be `'KL'` or `'KS'`
     - `l1` and `l2`: should be `'e'` or `'mu'`
     """
     # masses
@@ -37,6 +35,40 @@ def amplitudes(par, wc, l1, l2):
     return xi_t * P, xi_t * S
 
 
+def amplitudes(par, wc, K, l1, l2):
+    r"""Amplitudes P and S entering the $K_{L,S}\to\ell_1^+\ell_2^-$ observables.
+    
+    Parameters
+    ----------
+    
+    - `par`: parameter dictionary
+    - `wc`: Wilson coefficient dictionary
+    - `K`: should be `'KL'` or `'KS'`
+    - `l1` and `l2`: should be `'e'` or `'mu'`
+    """
+    # KL, KS are linear combinations of K0, K0bar. So are the amplitudes.
+    S_K0bar, P_K0bar = amplitudes_weak_eigs(par, wc, l1, l2)
+    if l1 != l2:
+        S_aux, P_aux = amplitudes_weak_eigs(par, wc, l2, l1)
+        S_K0 = -S_aux.conjugate()
+        P_K0 = P_aux.conjugate()
+        if 'K' == 'KL':
+            sig = +1
+        elif 'K' == 'KS':
+            sig = -1
+        S = (S_K0 + sig * S_K0bar) / 2
+        P = (P_K0 + sig * P_K0bar) / 2
+    # Save computing time for special cases. See also arXiv:1711.11030.
+    elif l1 == l2:
+        if 'K' == 'KL':
+            S = -1j * S_K0bar.imag 
+            P = P_K0bar.real
+        elif 'K' == 'KS':
+            S = -S_K0bar.real
+            P = -1j * P_K0bar.imag
+    return P, S
+
+
 def amplitudes_LD(par, K, l):
     r"""Long-distance amplitudes entering the $K\to\ell^+\ell^-$ observables."""
     ml = par['m_' + l]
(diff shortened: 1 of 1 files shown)


peterstangl:
@hoodyn thank you very much for your PR and the detailed explanations of what you did! I hope I find time soon to check everything in detail.

flavio/physics/kdecays/kll.py
```diff
@@ -33,38 +31,73 @@ def amplitudes(par, wc, l1, l2):
...
+    - `K`: should be `'KL'` or `'KS'`
+    - `l1` and `l2`: should be `'e'` or `'mu'`
+    """
+    # KL, KS are linear combinations of K0, K0bar. So are the amplitudes.
+    S_K0bar, P_K0bar = amplitudes_weak_eigst(par, wc, l1, l2)
+    if l1 != l2:
+        S_aux, P_aux = amplitudes_weak_eigst(par, wc, l2, l1)
+        S_K0 = -S_aux.conjugate()
+        P_K0 = P_aux.conjugate()
+        if 'K' == 'KL':
```
peterstangl, line 56: I think here it should be
```python
if K == 'KL':
```
(you are currently comparing two strings and will always get a `False`)

flavio/physics/kdecays/kll.py
```diff
@@ -33,38 +31,73 @@ def amplitudes(par, wc, l1, l2):
...
+            sig = +1
+        elif 'K' == 'KS':
```
peterstangl, line 58: Same thing as above, it should be
```python
if K == 'KS':
```
(you are currently comparing two strings and will always get a `False`)

flavio/physics/kdecays/kll.py
```diff
@@ -33,38 +31,73 @@ def amplitudes(par, wc, l1, l2):
...
+            sig = -1
+        S = (S_K0 + sig * S_K0bar) / sqrt(2)
+        P = (P_K0 + sig * P_K0bar) / sqrt(2)
+    # Save computing time for special cases. See also arXiv:1711.11030.
+    elif l1 == l2:
+        if 'K' == 'KL':
```
peterstangl, line 64:
```python
if K == 'KL':
```

flavio/physics/kdecays/kll.py
```diff
@@ -33,38 +31,73 @@ def amplitudes(par, wc, l1, l2):
...
+            S = -1j * S_K0bar.imag * sqrt(2)
+            P = P_K0bar.real * sqrt(2)
+        elif 'K' == 'KS':
```
peterstangl, line 67:
```python
if K == 'KS':
```

peterstangl (changes requested):
I found some small typos

flavio/physics/kdecays/kll.py
hoodyn: Sure, you are right.

flavio/physics/kdecays/kll.py
hoodyn: Sure, sorry for the mess.

peterstangl (approved):

hoodyn:
@peterstangl  I'm attaching some notes which may help you checking the code after me.
 
[kaon_decays.pdf](https://github.com/flav-io/flavio/files/5894970/kaon_decays.pdf)

hoodyn:
For too many bugs and misargumentation in this PR, I've iniciated a new one (#141).

PR #141, opened 2021-02-10
Fixing calculation of BR(KL,KS->emu,mue)

hoodyn:
This is the (second) pull request aiming to fix the problems in calculations od LFV kaon decays to leptons, based on discussion in #138. (I'm aborting my original PR #140 which I polluted by bugs and wrong argumentation.)

### DIFF
--- a/flavio/physics/kdecays/kll.py
+++ b/flavio/physics/kdecays/kll.py
@@ -9,8 +9,8 @@ from flavio.physics.common import add_dict
 from math import pi, sqrt
 
 
-def amplitudes(par, wc, l1, l2):
-    r"""Amplitudes P and S entering the $K\to\ell_1^+\ell_2^-$ observables.
+def amplitudes_weak_eigst(par, wc, l1, l2):
+    r"""Amplitudes P and S for the decay $\bar K^0\to\ell_1^+\ell_2^-$.
 
     Parameters
     ----------
@@ -37,6 +37,41 @@ def amplitudes(par, wc, l1, l2):
     return xi_t * P, xi_t * S
 
 
+def amplitudes(par, wc, K, l1, l2):
+    r"""Amplitudes P and S entering the $K_{L,S}\to\ell_1^+\ell_2^-$ observables.
+    
+    Parameters
+    ----------
+    
+    - `par`: parameter dictionary
+    - `wc`: Wilson coefficient dictionary
+    - `K`: should be `'KL'` or `'KS'`
+    - `l1` and `l2`: should be `'e'` or `'mu'`
+    """
+    # KL, KS are linear combinations of K0, K0bar. So are the amplitudes.
+    # Normalization differs by sqrt(2) from `amplitudes_weak_eigst`.
+    P_K0bar, S_K0bar = amplitudes_weak_eigst(par, wc, l1, l2)
+    if l1 != l2:
+        P_aux, S_aux = amplitudes_weak_eigst(par, wc, l2, l1)
+        S_K0 = -S_aux.conjugate()
+        P_K0 = P_aux.conjugate()
+        if K == 'KL':
+            sig = +1
+        elif K == 'KS':
+            sig = -1
+        S = (S_K0 + sig * S_K0bar) / 2
+        P = (P_K0 + sig * P_K0bar) / 2
+    # Simplified expressions for special cases. See also arXiv:1711.11030.
+    elif l1 == l2:
+        if K == 'KL':
+            S = -1j * S_K0bar.imag 
+            P = P_K0bar.real
+        elif K == 'KS':
+            S = -S_K0bar.real
+            P = -1j * P_K0bar.imag
+    return P, S
+
+
 def amplitudes_LD(par, K, l):
     r"""Long-distance amplitudes entering the $K\to\ell^+\ell^-$ observables."""
     ml = par['m_' + l]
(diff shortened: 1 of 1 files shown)


hoodyn:
Summary of the issues being solved and proposed changes:
**Issues:**
1: Call for a prediction of `'BR(KS->emu,mue)'` was giving an error due to a bad implementation.
2: `BR(KL->emu,mue)` was calculated incorrectly, since simplified formulas were used which hold only for LF conserving cases.

**Suggested updates**
1. `K='KS', l1='e', l2='mu'` is no longer an invalid combination of options in the function `amplitudes_eff`.

2. Amplitudes for LFV decays (i.e. `l1!=l2`) were not calculated correctly. Fixing the physics required some nontrivial extension of the code:

- the original function function `amplitudes` has been renamed to `amplitudes_weak_eigst` as it actually does not yield the relevant amplitudes for _KL -> l1+l2-_ nor _KS->l1+l2-_, but for the decay of weak eigenstate _K0bar -> l1+ l2-_. Thus the usage:
```
S_K0bar, P_K0bar = amplitudes_weak_eigst(par, wc, l1, l2)
```
Also the amplitude for K0->l1+l2- can be obtained from this function as
```
S_aux, P_aux = amplitudes_weak_eigst(par, wc, l2, l1)
S_K0 = -S_aux.conjugate()
P_K0 = P_aux.conjugate()
```
I will attach some updated notes explaining these matters in detail.

- An entirely new function `amplitudes(par, wc, K, l1, l2)` has been implemented. It simply makes proper linear combinations of amplitudes for K0 and K0bar which correspond to the approximate mass eigenstates KL, KS (in fact, CP eigenstates are used instead). In the special case `l1 == l2`, this can be simplified to the expressions with imaginary or real parts of S,P, which had been there originally, adopted from ArXiv:1711.11030 (I guess). I implemented this special case separately inside the amplitudes, for the reason of both effectivity and better possibility to check that the new formulae are consistent with the old ones. If the maintainers find this separate treatment redundant, it can be removed, of course. The new amplitudes function now takes `K` as a mandatory argument (to be `'KL'` or `'KS'`).
- There is no longer a need to take real or imaginary parts of the amplitudes within the `amplitudes_eff` function, which now serves purely to sum the short-distance `amplitudes` with `amplitudes_LD` in the LF-conserving cases. It also seems more consistent now to me. **Concerning LF-conserving decays:** Due to different sign conventions, `amplitudes_eff` _subtracts_ the LD and SD contributions in this version, but in fact the interference works the same way as it did before. (For `'KL'`, an extra minus sign had been present in original implementation of `P` in `amplitudes_LD`. For `'KS'`, the sign is compensated with `S=-S_K0bar.real` inside  `amplitudes` function yields .

Furthermore, I made an additional small commit:
3. The function `amplitudes_LD(par, K, l)` was originally giving long-distance contributions for both `'KL'` and `'KS'`, although it takes `K` as an argument. Eventually, everything had been calculated correctly but the implementation seemed little awkward to me. So I've changed it in order to make things clearer.

hoodyn:
Here's my notes about the leptonic kaon decays in EFT. (I haven't found any suitable text in the literature.) Might be helpful to understand the changes in the implementation.
[kaon_decays.pdf](https://github.com/flav-io/flavio/files/5963600/kaon_decays.pdf)

peterstangl:
@hoodyn thank you again for all the work you have put into this! Since your results differ from those in published papers, I think it would be good to do some cross checks before merging this PR. Unfortunately, I don't have enough time at the moment to do that. So I will keep this PR open until I find time for this or somebody else can do some cross checks. Did you actually contact the authors of the paper that seems to contain mistakes?

hoodyn:
@peterstangl 
Hi, I haven't contacted the authors of 1602.00881. I think that the mistakes in that paper can be qualified just as "big typos". This paper does not give formulae for KL,KS -> emu anyway, so I don't think it is important for this PR. 

Btw, I used my branch of flavio to successfully crosscheck predictions of Ref. 1801.02895. The same crosscheck fails for the current master branch. This, however, says nothing about the problems I see in Ref. 1602.00881.

> @hoodyn thank you again for all the work you have put into this! Since your results differ from those in published papers, I think it would be good to do some cross checks before merging this PR. Unfortunately, I don't have enough time at the moment to do that. So I will keep this PR open until I find time for this or somebody else can do some cross checks. Did you actually contact the authors of the paper that seems to contain mistakes?

peterstangl:
> Here's my notes about the leptonic kaon decays in EFT. (I haven'tfound any suitable text in the literature.) Might be helpful to
understand the changes in the implementation.
[kaon_decays.pdf](https://github.com/flav-io/flavio/files/5963600/kaon_decays.pdf)

@hoodyn I cross-checked your notes and found overall agreement. I found
a few disagreements, which are, however, probably just typos. Nevertheless, I list them in the following:
- In Eq. (16) I think there is a factor `m_K0` missing.
- In Eq. (20) I think there should be a minus instead of a plus inside the square
bracket on the second line (`S_sdmue S_sdemu - P_sdmue P_sdemu`).
- In Eqs. (25a-25d) I think \bar{m}_K should be replaced by \bar{m}_K/m_K and m_mu should be replaced by m_l everywhere.

Concerning the implementation, I will comment directly on the code.

flavio/physics/kdecays/kll.py
```diff
@@ -33,52 +32,92 @@ def amplitudes(par, wc, l1, l2):
...
+        if K == 'KL':
+            sig = +1
+        elif K == 'KS':
+            sig = -1
+        S = (S_K0 + sig * S_K0bar) / 2
+        P = (P_K0 + sig * P_K0bar) / 2
+    # Simplified expressions for special cases. See also arXiv:1711.11030.
+    elif l1 == l2:
+        if K == 'KL':
+            S = -1j * S_K0bar.imag
```
peterstangl, line 67: comparing this to your notes, Eq. (22a), shouldn't this be `+1j * S_K0bar.imag`?

flavio/physics/kdecays/kll.py
```diff
@@ -33,52 +32,92 @@ def amplitudes(par, wc, l1, l2):
...
+    Parameters
+    ----------
+    
+    - `par`: parameter dictionary
+    - `wc`: Wilson coefficient dictionary
+    - `K`: should be `'KL'` or `'KS'`
+    - `l1` and `l2`: should be `'e'` or `'mu'`
+    """
+    # KL, KS are linear combinations of K0, K0bar. So are the amplitudes.
+    # Normalization differs by sqrt(2) from `amplitudes_weak_eigst`.
```
peterstangl, line 52: Is there a particular reason not to use the same normalization?

flavio/physics/kdecays/kll.py
```diff
@@ -33,52 +32,92 @@ def amplitudes(par, wc, l1, l2):
...
         PLD = 0
     else:
         SLD, PLD = amplitudes_LD(par, K, l1)
-    if K == 'KS' and l1 == l2:
-        Peff = P.imag
-        Seff = S.real + SLD
-    if K == 'KL':
-        Peff = P.real + PLD
-        Seff = S.imag
+    # The relative sign due to different conventions. Cf. my notes.
```
peterstangl, line 101: Maybe instead of `Cf. my notes.`, you could put the link to the notes there. Otherwise it will be impossible find them just by looking at this comment in the code.

PR #142, opened 2021-02-16
Remove matplotlib deprecation warnings

MJKirk:
Removes a slightly annoying deprecation warning.
Setting the rcparam as a list rather than a single string has been deprecated in matplotlib v3.3.0 (see https://matplotlib.org/3.3.0/api/api_changes.html#setting-rcparams-text-latex-preamble-default-or-rcparams-pdf-preamble-to-non-strings)


peterstangl:
Great, thanks a lot!

PR #146, opened 2021-03-01
add observable BR(B0->K*gamma)/BR(Bs->phigamma)

peterstangl:


flavio/physics/bdecays/bvgamma.py
```diff
@@ -179,6 +185,23 @@ def BVgamma_function(function, B, V):
...
+_process_den_tex = r"B_s\to \phi\gamma"
+_obs.set_description(
+    fr"Ratio of branching ratio of ${_process_num_tex}$ and time-integrated "
+    fr"branching ratio of ${_process_den_tex}$"
+)
+_obs.tex = (
+    fr"$\frac{{\text{{BR}}({_process_num_tex})}}"
+    fr"{{\overline{{\text{{BR}}}}({_process_den_tex})}}$"
+)
+_obs.add_taxonomy(_process_taxonomy + _process_den_tex + r"$")
```
peterstangl, line 200: @DavidMStraub for the taxonomy I used `B_s\to \phi\gamma`. Do you think it would be better if I add a new one just for this observable?

DavidMStraub: Sorry, missed this comment. I'd suggest to use both of them (so the observable is listed under phigamma *and* Kstargamma in the docs).

DavidMStraub: Like e.g. here:
https://github.com/flav-io/flavio/blob/ea735bd8febbb961d249eddf338a4960c1fbee69/flavio/physics/wdecays/gammaw.py#L115-L117

peterstangl: Thanks, that's a very good idea! I've updated the taxonomy in https://github.com/flav-io/flavio/commit/b99c2ae3cd08c1e6440e305687c54bbf3d77f1c7.

PR #150, opened 2021-03-08
Add B->ee measurements

hoodyn:
Hi, I would like to add the recent LHCb limits on _B0 -> e+e-_ and _Bs -> e+e-_ to flavio. These processes are important in vector LQ scenarios, for example. There has been no other measurement of these decays included in flavio so far.


peterstangl:
Thanks @hoodyn!

PR #155, opened 2021-03-31
Add inspire to 'LHCb KS->mumu 2019' measurement.

hoodyn:
The measurement was only released in a preliminary report (inspire: LHCb:2019aoh) when implemented to flavio but in the meantime it has been normally published (inspire: Aaij:2020sbt).

### DIFF
--- a/flavio/data/measurements.yml
+++ b/flavio/data/measurements.yml
@@ -7053,7 +7053,7 @@ LHCb KS->mumu 2019:
   experiment: LHCb
   inspire: Aaij:2020sbt
   values:
-    BR(KS->mumu): 0.94 +0.72 -0.64 e-10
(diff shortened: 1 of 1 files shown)


peterstangl:
Thank you @hoodyn! I saw that in the published version the numbers have slightly changed (they have one significant digit less). So maybe it might be good to adjust them also in flavio in order to be in agreement with the published version. It would be great if you could include these small modifications into the PR before we merge it.

hoodyn:
> Thank you @hoodyn! I saw that in the published version the numbers have slightly changed (they have one significant digit less). So maybe it might be good to adjust them also in flavio in order to be in agreement with the published version. It would be great if you could include these small modifications into the PR before we merge it.

Done.

peterstangl:
Great, thanks @hoodyn!

Issue #158, opened 2021-05-07
Taxonomy of R_uc is wrong


MJKirk:
It's listed under Z->bb (https://flav-io.github.io/docs/obs/zdecays-flavourconservingdecays.html#z0to-bbar-b) rather than Z->uu and Z->cc.
This is because `_process_tex` in 
https://github.com/flav-io/flavio/blob/1546772eefabfb8abbfe2c550faa1957cfadae32/flavio/physics/zdecays/gammaz.py#L138
isn't set, and is just carrying the last value from the loop a few lines above.
I'm busy right now else this would have just been a pull request, but I can do one whenever I get the chance.

PR #159, opened 2021-05-11
Fixes taxonomy of R_uc observable

MJKirk:
_process_tex wasn't set, so it was still set to the last value from the loop above and hence the taxonomy was wrong.
(As described in #158)
The R_uc observable will now be listed under both Z->uu and Z->cc.


peterstangl:
Great, thanks a lot @MJKirk!

Issue #156, opened 2021-04-09
Slightly larger SM uncertainties for RD(*) compared to literature


Jorge-Alda:
For a long time we have been using version 1.5.0, which produces a SM uncertainty for the observables Rtaul(B->Dlnu) and Rtaul(B->D*lnu) of 0.006 and 0.003 respectively, which are similar to the values published by HFLAV (https://hflav-eos.web.cern.ch/hflav-eos/semi/spring19/html/RDsDsstar/RDRDs.html).

However, we recently updated Flavio to version 2.2.0, and we have found that the SM uncertainties calculated by this new version are much larger, 0.087 and 0.031 respectively. If these values were true, the tension in the RD(\*) observables would disappear (see attached figure).

We have examined the source code, and the functions used to compute RD(\*) or the SM uncertainty have not changed since version 1.5.0. The only relevant change, as far as we know, could be the new B->D and B->D\* form factors introduced in version 2.0.

Is this an intended behaviour of the code? Is there any way to produce uncertinty predictions similar to those of HFLAV in the newer versions of Flavio?

[RDplots.pdf](https://github.com/flav-io/flavio/files/6283907/RDplots.pdf)

peterstangl:
@Jorge-Alda thank you for reporting this. From `flavio.sm_error_budget('Rtaul(B->Dlnu)')` and `flavio.sm_error_budget('Rtaul(B->D*lnu)')` one indeed finds that there are the entries
```python
('CLN lp_3(1)',
  'CLN c_xi',
  'CLN lp_6(1)',
  'chi_2p(1)',
  'eta(1)',
  'chi_2pp(1)',
  'CLN l_4(1)',
  'CLN l_6(1)',
  'etap(1)',
  'CLN l_2(1)',
  'CLN lp_4(1)',
  'CLN xi3',
  'etapp(1)',
  'chi_3p(1)',
  'CLN l_3(1)',
  'CLN lp_5(1)',
  'CLN lp_2(1)',
  'CLN rho2_xi',
  'chi_2(1)',
  'chi_3pp(1)',
  'CLN l_1(1)',
  'CLN lp_1(1)',
  'CLN l_5(1)'): 0.32750804819346513,
```
and
```python
('CLN lp_3(1)',
  'CLN lp_6(1)',
  'CLN c_xi',
  'chi_2p(1)',
  'eta(1)',
  'chi_2pp(1)',
  'CLN l_4(1)',
  'CLN l_6(1)',
  'etap(1)',
[...]

PR #160, opened 2021-05-25
Fixed wrong signs in the correlation matrix of the CLN parameters

Jorge-Alda:
Fixed the wrong signs that are causing #156


peterstangl:
Thank you @Jorge-Alda!

PR #162, opened 2021-06-01
fixed handling of subleading contributions to B->Vll decays at high q2

langenbruch:
Fixes the handling of the subleading contributions to the B->Vll decay amplitudes, as `helicity_amps_v` expects the vector coupling in `wc['v']`, not in `wc['9']`.


peterstangl:
Thanks a lot @langenbruch!

PR #164, opened 2021-06-18
Removal of superfluous factors of 1/2 from the conversion of Wilson coefficients of (pseudo)scalar operators in the function get_wceff_fccc

whereforebound:
We had looked at the contributions of the scalar operator to the lepton-flavour universality ratios R(D) and R(D*) and noticed a discrepancy between the results output by flavio and those obtained from analytic expressions in the literature.
In the function get_wceff_fccc, there is a factor 1/2 involved in the conversion of the Wilson coefficients of the relevant scalar operators to those used for the helicity amplitudes in flavio which appears superfluous.
After removing this factor, the flavio results for R(D) and R(D*) are compatible with the literature ones. This is illustrated by the two attached plots. They show the results for R(D)/R(D)_{SM} and R(D*)/R(D*)_{SM} as functions of the scalar Wilson coefficient C_{SL} obtained via flavio and from analytic expressions used in several papers referenced in the plot legend, with the aforementioned factor of 1/2 included ("_old.png") and removed ("_new.png"), respectively. In order to exclude the possibility of RG effects playing a role, C_{SL} is defined at the bottom-quark mass scale. Furthermore, C_{SL} is chosen real and all contributions from other operators are switched off.

R(D*) without factor 1/2:

scalaroperatorRDStar_new

R(D*) with factor 1/2:

scalaroperatorRDStar_old

R(D) without factor 1/2:

scalaroperatorRD_new

R(D) with factor 1/2:

scalaroperatorRD_old

### DIFF
--- a/flavio/physics/bdecays/wilsoncoefficients.py
+++ b/flavio/physics/bdecays/wilsoncoefficients.py
@@ -216,10 +216,10 @@ def get_wceff_fccc(wc_obj, par, qiqj, lep, nu, mqi, scale, nf=5):
     c['vp'] = wc['CVR_'+qqlnu]/2.
     c['a']  = -(c_sm + wc['CVL_'+qqlnu])/2.
     c['ap'] = -wc['CVR_'+qqlnu]/2.
-    c['s']  = 1/2 * wc['CSR_'+qqlnu]/2.
-    c['sp'] = 1/2 * wc['CSL_'+qqlnu]/2.
-    c['p']  = -1/2 * wc['CSR_'+qqlnu]/2.
-    c['pp'] = -1/2 * wc['CSL_'+qqlnu]/2.
+    c['s']  = wc['CSR_'+qqlnu]/2.
+    c['sp'] = wc['CSL_'+qqlnu]/2.
+    c['p']  = -wc['CSR_'+qqlnu]/2.
+    c['pp'] = -wc['CSL_'+qqlnu]/2.
     c['t']  = 0
     c['tp'] = wc['CT_'+qqlnu]
     return c


DavidMStraub:
I would be very surprised if this were the case. I am pretty sure this was compared e.g. to @MartinJung1978's independent implementation. Are you sure it's not an Operator definition issue or something like that?

micha-a-schmidt:
Looking at the definition of the scalar helicity amplitude H['s'] in the function helicity_amps_v in angular.py, the definition of the Wilson coefficients seems to agree with 1506.03970, where O_S=\bar s P_R b \bar l l and O_V=\bar s \gamma^\mu P_L b \bar l \gamma^\mu l, i.e. there is a projection operator for both quark currents, but there are no projection operators for the lepton currents. The weak effective theory (WET) basis used in flavio uses projection operators for both quark and lepton currents for the relevant operators of b->clnu, i.e. to convert from the flavio WET to the basis used in 1506.03970, I have to divide both vector and scalar Wilson coefficients by 2. 

However, there is an additional factor of 1/2 for the scalar Wilson coefficients in the function get_wceff_fccc: e.g. line 219 reads
c['s']  = **1/2** * wc['CSR_'+qqlnu] **/2.**
similarly the ones for c['sp'], c['p'] and c['pp'] in the following lines.

Comparing with the neutral current process b->pll, where the translation of the Wilson coefficients is implemented in get_wceff, as far as I can see, there is no relative factor of 1/2 for the scalar Wilson coefficients, when translating from C9_bsmumu and CS_bsmumu to the Wilson coefficients used in 1506.03970. Also for the LFV decay b->pll which uses get_wceff_lfv, there is no relative factor of 1/2 in the code. As expected @whereforebound finds that the result agrees well with 2106.05192.

Thus we suspect that the additional factors of 1/2 for the scalar Wilson coefficients in get_wceff_fccc should be removed.

DavidMStraub:
I don't have time to look into this in detail. @MartinJung1978 perhaps you remember whether we compared the scalar contributions to B->D*lnu or whether there could be a wrong factor of 2 in your opinion.

MartinJung1978:
I remember that we checked all NP contributions with my independent implementation, we had agreement. We found an issue with the tensors in one of the sets of expressions in Datta et al., but the (pseudo-)scalar contributions were fine. I also compared with my earlier work on scalars in b->ctaunu and found agreement, that was a third implementation.

DavidMStraub:
Thanks!

DavidMStraub:
I'd be convinced there is an error if EOS disagress with flavio, too, using WCxf input.

peterstangl:
I have added the EOS prediction to the notebook provided by @micha-a-schmidt and it seems like currently there is indeed a factor of 2 missing in flavio: https://gist.github.com/peterstangl/8669ac414d0468b856553d080b27838d
(WCxf input in EOS did not work and I will open an issue in the EOS repository)
@MartinJung1978 in your independent implementation, what do you find for R(D)/R(D)_SM for `CSL_bctaunutau = 0.3`? In flavio it currently is 1.26, while EOS finds 1.56.

peterstangl:
I went through the operator definitions and also from that I find agreement with the changes made in this PR.

peterstangl:
I think this has been checked now and can be merged.

DavidMStraub:
Great work everybody, I guess I screwed this up.

Issue #171, opened 2021-11-26
fK form factor correlation with fpi


MJKirk:
It's not clear to me where the flavio value of the correlation between `f_K` and `f_pi` comes from.
Reading the comments in
https://github.com/flav-io/flavio/blob/3b8032913b79d01a48190ca7ab3366ff67b70173/flavio/data/parameters_correlated.yml#L26-L41
it suggests it comes from FLAG, but I can't see that value in the review.
Alternatively if it comes from reverse engineering it to reproduce the ratio `f_K/f_pi` as suggested by @DavidMStraub's earlier commit (https://github.com/flav-io/flavio/commit/0f75d69eed5746b8d939a0140755f40a1b20dbbd), then it seems to me that both the error on `f_K` and the correlation between `f_K` and `f_pi` are not known, and there is a whole range of values for both that would reproduce the FLAG error on the ratio.
E.g `f_K = 0.1554(8)` and a correlation of `0.975` would also reproduce the ratio correctly.

DavidMStraub:
Which is the value you can't see in the review? From table 1:

<img width="540" alt="grafik" src="https://user-images.githubusercontent.com/10965193/143603334-8e9afe51-40bb-4b09-b39c-b2125f83a754.png">

MJKirk:
The correlation coefficient between fK and fpi.
I'm saying it seems like you took fK/fpi and fpi from FLAG, and then reverse engineered **both** fK and the correlation between fK and fpi. But I don't see how there is enough information for that. Unless the correlation is already known

DavidMStraub:
I see your point. The central value for fK obviously comes from the ratio, but the correlation can only be extracted if the uncertainty on fK is known, and there is no reference given for that. No idea where it came from. But I suspect there was a perfectly well thought through & legitimate reasoning behind it :thinking: 

What's clear is that the 0.3 MeV error (2+1+1) is not applicable as it uses the experimental value of fpi to set the lattice scale.

MJKirk:
Okay, glad I have not missed something.
It appears from some quick calculations that if you just take both the uncertainty and the correlation as free parameters, then that particular value for the fK uncertainty minimises the correlation you need to reproduce fK/fpi. So maybe 2019 David thought that was a good reason?

DavidMStraub:
Fortunately 2019 David didn't throw away his data.

I found the notebook: https://gist.github.com/DavidMStraub/813bd0c4296837936a188a29c8d8b981

[...]

PR #173, opened 2022-01-10
Improved comment on fK and meson mixing

MJKirk:
- Added a comment to the parameter file to summarise my discussion with David in #171 
- Add a note that the meson mixing matrix element values have the 1/2M normalisation included - this is something I periodically forget in flavio, and spend a while wondering why the code is  `M * f^2 * bag` not `M^2 * f^2 * bag` before remembering. I wasn't sure whether to just add a comment in the code or alter the docstring - I decided on the docstring since someone (possibly future me again) might look there first after being confused about the values that `matrixelements` returns before looking at the source code.


peterstangl:
Thanks @MJKirk! I think it makes sense to have the comment in the docstring.

PR #177, opened 2022-02-13
added LHCb R(KS) and R(K*+) measurement

nsahoo:
Added the recent LHCb measurements R(KS) and R(K*+) from 2021. 
Reference: https://arxiv.org/pdf/2110.09501.pdf
@peterstangl: please merge it after reviewing.


peterstangl:
Thank you @nsahoo for this PR! Since the profile likelihoods for both measurements are available, I think it would be better to actually add those. This has also been done for the other LHCb measurements of RK and RK*. Do you think you could add a numerical likelihood as done e.g. here?:
https://github.com/flav-io/flavio/blob/b7ac308ac83cd7168d80c3382b5d65415bc64679/flavio/data/measurements.yml#L9985-L10028

nsahoo:
@peterstangl this may sound stupid, yet let me ask: how to extract these numerical values ?

peterstangl:
> @peterstangl this may sound stupid, yet let me ask: how to extract these numerical values ?

That's not a stupid question since in the past, this was indeed rather complicated and often required digitizing the plots. But fortunately, this has changed and the plot data is now provided by LHCb as "Related data files" on the CERN Document Server: http://cds.cern.ch/record/2784954/files/

nsahoo:
@peterstangl great, thanks for pointing me out. I was looking for hepdata, but didn't get and that's why asked. I'll have a look at the supplementary files for these data points. 
cheers, Niladri.

nsahoo:
Thanks a lot @harryvcliff for providing these numerical likelihoods.

DavidMStraub:
If you look at the Travis log, you will see that it complains about a syntax error in your YAML. Since this YAML is parsed when importing the package, all tests fail.

nsahoo:
@DavidMStraub thanks a lot for your reply, I still don't get where I did wrong so that it's still throwing the same syntax error.

peterstangl:
Hi @nsahoo,

`distribution: numerical` as well as the lines containing `x:` and `y:` have to be indented with respect to `value:` (see the example from the RK measurement in my comment above).

One question: did you actually implement the likelihoods for R(KS) and R(K*+) or for R(KS)^-1 and R(K*+)^-1? The latter are shown in the supplementary material of the LHCb publication, but maybe @harryvcliff provided you with the former? As you probably noticed, the R(KS)^-1 and R(K*+)^-1 observables are not defined in flavio at the moment. Would it actually be better to use them (since this was also the choice of LHCb in their publication)? Then we could easily add them.

nsahoo:
Hi @peterstangl 

The likelihoods for R(KS) and R(K*+) are provided by @harryvcliff, I put it accordingly just to be aligned with other available LFU measurements (in the same way defined in flavio).

peterstangl:
@nsahoo I took a closer look at why the checks fail and it seems that you used a tab character for indentation, which, however, is not allowed by YAML syntax.

Concerning R vs. R^-1, the LHCb paper says "The inverse ratio R^-1 is measured as its uncertainty better represents a Gaussian distribution due to the low yield of the electron decay mode". So this means that for some applications it could actually be advantageous to have the R^-1 measurements implemented as well. But I would suggest we first use your PR to implement the R measurements and then later when R^-1 will be available as an observable in flavio, one can implement these measurements as well. And then the user could choose which one to use.

Regarding R(pK), the measurement is not included because there is no theoretical prediction available at the moment. It might be possible to implement something along the lines of https://arxiv.org/abs/2110.09882.

nsahoo:
hi @peterstangl 

thanks a lot for your detailed comments and my apologies for this stupid indentation issue. Can it be merged now (if there are no further issues) ?

peterstangl:
Thank you @nsahoo for the update! Unfortunately, the automatic tests now stopped working and we have to setup a GitHub Actions workflow to make them work again. I hope to find time for doing this soon.

nsahoo:
no issues @peterstangl

peterstangl:
Hi @nsahoo, thanks to @DavidMStraub the GitHub Actions workflow is now set up and we can run tests again! I noticed that you still have tab characters as indentation inside the `x` and `y` arrays. Could you please change them to normal spaces and ideally align each line with the first entry after the initial `[`? I.e. e.g.
```yaml
        x: [0.3333, 0.3336, 0.3339, 0.3343, 0.3346, 0.3349, 0.3352, 0.3355, 0.3358, 0.3361, 0.3364, 0.3367, 0.3370, 0.3374, 0.3377, 0.3380, 0.3383, 0.3386,
            0.3389, ...
```
As soon as you have changed this and pushed your new commit, the GitHub Actions workflow will be triggered and will run all tests. Thanks!

peterstangl:
@nsahoo the tests fail since there are still tab characters around that prevent the YAML import. This time it is at the end of line 10175 after the url:
```
E   yaml.scanner.ScannerError: while scanning for the next token
E   found character '\t' that cannot start any token
E     in "<byte string>", line 10175, column 95:
E        ... tPublic/LHCb-PAPER-2021-038.html	
E                                            ^
```
Maybe you can search for tab characters in the file and remove all of them. You can then also run the tests locally using for example `nosetests` or `pytest` in the terminal, or you can just try to `import flavio` before pushing your changes.

DavidMStraub:
I can also recommend using VS Code with automatic formatting (e.g. for YAML files).

nsahoo:
hi @peterstangl, @DavidMStraub 

Thanks a lot for your comments/suggestions, I did the local tests and it showed few warning messages only. It should be fine to run the GitHub Actions workflow now.

peterstangl:
OK, now all the checks pass. Thank you @nsahoo!

PR #179, opened 2022-02-22
Add workflow

DavidMStraub:
This adds a basic unit test workflow.

I did not add coveralls yet - don't know how.

Using pytest instead of nose2, can be changed.


peterstangl:
That's great! Thanks a lot @DavidMStraub! We can add coveralls back later.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Python dependencies
+        run: |
+          pip install .[sampling,plotting,testing]
+          pip install coveralls
```
peterstangl, line 24: We could remove this line for now since coveralls doesn't work at the moment.

peterstangl:
@DavidMStraub, it looked quite good, but unfortunately, the tests are failing for different reasons: https://github.com/flav-io/flavio/actions/runs/1883183194

For Python 3.7 and 3.9 installation of flavio works but `pytest` is not found.
The installation for 3.8 has been cancelled and I don't know why and 3.10 doesn't install because of https://github.com/DavidMStraub/rundec-python/issues/4

DavidMStraub:
I'll take care of it, now that it runs it should be straightforward to fix/improve.

DavidMStraub:
> The installation for 3.8 has been cancelled and I don't know why

Github actions usually cancels all jobs as soon as one of them fails.

Issue #172, opened 2021-12-13
Bug in epsilon_K calculation


MJKirk:
In the meson mixing code, the wilson coefficients (specifically the loop functions) use the quark masses at a single fixed scale.
https://github.com/flav-io/flavio/blob/3b8032913b79d01a48190ca7ab3366ff67b70173/flavio/physics/mesonmixing/wilsoncoefficient.py#L59-L61
However, these loop functions get multiplied by the QCD RG correction factors
https://github.com/flav-io/flavio/blob/3b8032913b79d01a48190ca7ab3366ff67b70173/flavio/physics/mesonmixing/amplitude.py#L52-L57
where the values used assume the quark masses are mt(mt) and mc(mc) (see the first page of [1108.2036](https://arxiv.org/abs/1108.2036) for `eta_cc` or section 2 of [1007.0684](https://arxiv.org/abs/1007.0684) for `etc_ct`).

For the B mixing observables it doesn't make a difference, since they only use the `tt` coefficient, and mt(160 GeV) is very close to mt(mt).
But for epsilon_K, the difference between mc(160 GeV) and mc(mc) is large, and gives a 20% difference to the numerical value of `eps_K`.

Since the flavio paramter `m_c` is already mc(mc) from the PDG, that is easy to fix - just use `mc = par["m_c"]`
The up mass doesn't matter, since it is so small - I would again just use the PDG value `par["m_u"]` which is the MSbar mass at 2 GeV.
For the top, as I said before numerically it doesn't matter. But for correctness I think the easiest thing to do would be add a `get_mt_mt` function to the running module. RunDec has a function (`mOS2mSI`) to do this, so we can just use that.

I can submit a pull request if you agree with the issue.

peterstangl:
Hi @MJKirk, thanks for reporting this! I have to take a closer look at the papers. But I am wondering if, when updating the implementation of epsilon_K, we should directly implement the more recent and more precise results from https://arxiv.org/abs/1911.06822. What do you think?

MJKirk:
Yes probably. But that's more than a quick 3 line fix :laughing:

I would also note that using the current (but fixed) flavio implementation and similar inputs to that paper, you get an uncertainty of about 0.23, close to their 0.18, so I think it would be reasonable to do the quick fix now so flavio is at least correct (if a bit less precise). And then work on the newer formula next

MJKirk:
[...]

PR #181, opened 2022-04-05
Simple fix for eps_K

MJKirk:
Since it's already been over two months since #172, and I haven't had time to even start the alternative implementation of epsK from https://arxiv.org/abs/1911.06822, I'm just submitting my quick fix here.

As discussed in #172, the RG factors assume the quark masses used are mq(mq), not mq(160 GeV).
- For the top, we add a new function to the running module get_mt_mt which calculates mt(mt) using the RunDec function mOS2mSI and the top pole mass.
- For the charm, the input parameter m_c is already mc(mc) from PDG, so use that.
- For the up, the input m_u is actually mu(2 GeV), but this is closer to mu(mu) than mu(160 GeV), and numerically the up mass is extremely unimportant anyway.

The epsK NP test has to be changed, since with the correct value of mc the charm contribution to imag(M12) is larger (around 30% vs 10% with the old value), and so the NP doesn't exactly double the SM value.

### DIFF
--- a/flavio/physics/mesonmixing/test_mesonmixing.py
+++ b/flavio/physics/mesonmixing/test_mesonmixing.py
@@ -190,4 +190,4 @@ class TestMesonMixing(unittest.TestCase):
         self.assertAlmostEqual(
             flavio.np_prediction('eps_K', w) / flavio.sm_prediction('eps_K'),
             2,
-            delta=0.15)  # difference due to NNLO evolution of SM contribution + charm contribution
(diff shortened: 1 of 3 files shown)


peterstangl:
I finally found some time to take a closer look and I agree with everything! :) Thanks a lot @MJKirk!

peterstangl:
I finally found some time to take a closer look and I agree with everything! :) Thanks a lot @MJKirk!

PR #183, opened 2022-04-21
Fixes contour label in matplotlib >= 3.5.0

### LINKED ISSUE #178: Contour plot labels with matplotlib >= 3.5.0
When using a version of matplotlib >= 3.5.0, the contour plot legends from flavio don't show up correctly (see attached photos)

v3 4 3
v3 5 0

I suspect the relevant change is that in v3.4.3, contour returned a `LineCollection` but now returns a `PathCollection` (see https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.5.0.html?highlight=pathcollection#contourset-always-use-pathcollection), and so `set_label` as called here:
https://github.com/flav-io/flavio/blob/b7ac308ac83cd7168d80c3382b5d65415bc64679/flavio/plots/plotfunctions.py#L698
has different behaviour, but that's as far as I got investigating (for now at least - happy to keep digging around).

MJKirk:
Fixes #178

### DIFF
--- a/flavio/plots/plotfunctions.py
+++ b/flavio/plots/plotfunctions.py
@@ -695,7 +695,9 @@ def contour(x, y, z, levels, *, z_min=None,
         ax.contourf(x, y, z, levels=levelsf, **_contourf_args)
     CS = ax.contour(x, y, z, levels=levels, **_contour_args)
     if label is not None:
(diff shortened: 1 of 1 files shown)


peterstangl:
Thanks a lot @MJKirk! I somehow completely forgot to merge this...

PR #190, opened 2022-10-20
fix issues in Gamma distributions 

peterstangl:
This PR fixes several issues in the implementation of the Gamma distribution and related distributions:

- So far the support of the Gamma distribution has extended from 0 (potentially shifted by the `loc` parameter) to the x value where the CDF is 1-2e-9. For a sizable number of total counts, this range can be very large, which can make convolutions with such a distribution numerically unstable. However, in this case, the probability contained in the region of small x would be completely negligible since the distribution would be very close to a Gaussian distribution with a mode far away from 0.
The solution implemented in this PR is to use a support given by the region between the x values at which the CDF is larger than 1e-9 and smaller than 1-1e-9, with the additional requirement that the mode is within the support (e.g. if the number of counts is 0 and the mode is at 0, the lower boundary of the support will also be at 0).

- For `GeneralGammaDistributionPositive`, a warning is raised if the background variance is vanishing or very small. However, this was only done if the number of total counts was not equal to 0. Furthermore, a very small non-vanishing variance was kept finite. The latter point results in the issue that the convolution of a very narrow Gaussian with a Gamma distribution can become numerically unstable and then gives wrong results.
The solution of this PR is to set the background variance exactly to 0 if it is below a certain threshold (i.e. if the standard deviation of the Gaussian is at least 100 times smaller than the sqrt of the variance of the gamma distribution), in which case the convolution becomes the trivial convolution of the gamma distribution with a delta distribution. Furthermore, this treatment is extended to the case in which the number of total counts equals 0.

- For `GeneralGammaDistributionPositive`, the convolution of the Gamma distribution with a Gaussian is problematic in particular if the background variance is sizable and the number of total counts is very small or zero. In this case, the mode of the initial gamma distribution is shifted by the convolution from being close or equal to zero to larger values. This can even lead to cases where, after the convolution, values of x close or equal to zero get excluded by a certain number of sigmas even though the initial gamma distribution has its mode exactly at zero. So instead of having an upper bound on x as expected, one can end up with a distribution that strongly disfavors vanishing values of x. The reason for this is that by convoluting with a Gaussian, part of the probability content of the initial gamma distribution close to zero is moved to negative values, thereby reducing the probability of small positive values of x. This reduction remains after the final distribution is restricted to positive values of x.
The solution of this PR is to use a *folded normal distribution* (see e.g. https://en.wikipedia.org/wiki/Folded_normal_distribution) for the convolution instead of a normal distribution. The folded normal distribution is equal to a normal distribution for large positive values of x and becomes a half-normal distribution for x=0. Consequently, if the mode of the initial Gamma distribution is at x=0, it will stay there and the resulting distribution will still be an upper bound.
This procedure is implemented by first "mirroring" the gamma distribution at 0 (potentially shifted by the `loc` parameter), then convoluting the resulting distribution with a conventional normal distribution and then restricting the result to positive values of x.

### DIFF
--- a/flavio/statistics/probability.py
+++ b/flavio/statistics/probability.py
@@ -542,9 +542,10 @@ class GammaDistribution(ProbabilityDistribution):
         self.scipy_dist = scipy.stats.gamma(a=a, loc=loc, scale=scale)
         mode = loc + (a-1)*scale
         # support extends until the CDF is roughly "6 sigma"
-        support_limit = self.scipy_dist.ppf(1-2e-9)
+        support_min = min(self.scipy_dist.ppf(1e-9), mode)
+        support_max = self.scipy_dist.ppf(1-1e-9)
         super().__init__(central_value=mode, # the mode
-                         support=(loc, support_limit))
+                         support=(support_min, support_max))
         self.a = a
         self.loc = loc
         self.scale = scale
@@ -606,9 +607,10 @@ class GammaDistributionPositive(ProbabilityDistribution):
         if mode < 0:
             mode = 0
         # support extends until the CDF is roughly "6 sigma", assuming x>0
-        support_limit = self.scipy_dist.ppf(1-2e-9*(1-self.scipy_dist.cdf(0)))
+        support_min = max(min(self.scipy_dist.ppf(1e-9), mode), 0)
+        support_max = self.scipy_dist.ppf(1-1e-9*(1-self.scipy_dist.cdf(0)))
         super().__init__(central_value=mode, # the mode
-                         support=(0, support_limit))
+                         support=(support_min, support_max))
         self.a = a
         self.loc = loc
         self.scale = scale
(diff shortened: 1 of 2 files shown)


peterstangl:
@DavidMStraub if you have time it would be great if you can take a quick look at this PR :)

flavio/statistics/probability.py
```diff
@@ -542,9 +542,10 @@ def __init__(self, a, loc, scale):
         self.scipy_dist = scipy.stats.gamma(a=a, loc=loc, scale=scale)
         mode = loc + (a-1)*scale
         # support extends until the CDF is roughly "6 sigma"
-        support_limit = self.scipy_dist.ppf(1-2e-9)
+        support_min = min(self.scipy_dist.ppf(1e-9), mode)
```
DavidMStraub, line 545: Is `mode` guaranteed to be non-negative?

DavidMStraub: or should this be `loc` perhaps?

peterstangl: `mode` can be negative if it has been shifted by `loc` (see line 543). If the number of total events is 0 (i.e. `a`=1), then `mode`=`loc`. Otherwise `mode` is larger than `loc`.

DavidMStraub: I mean, does it make sense that `support_min` can end up negative?

peterstangl: I think for the gamma distribution with arbitrary `loc` parameter, a negative `support_min` should make sense. Note that in the previous implementation, `support_min` was actually always equal to `loc`, which is required to be negative or zero.

DavidMStraub: Makes sense :+1:

peterstangl:
@DavidMStraub did you actually also take a look at the second and third point described above? I think this PR should be fine and I will merge it soon.

peterstangl:
This PR has again been checked by @alekssmolkovic and myself. So I think it can be merged.

PR #191, opened 2022-10-28
Use Gaussian for measurement `LHCb B->ee 2020`

alekssmolkovic:
In order to use the measurement of `LHCb B->ee 2020` in the fast likelihood it should be Gaussian instead of an upper limit. This PR updates the measurement with the central values and uncertainties given by LHCb.


peterstangl:
Thanks for the PR!

PR #196, opened 2022-12-19
Assorted input updates

MJKirk:
This is most of the new inputs I made use of in my new paper on the CAA, plus just updating the version of PDG (I think the main thing here since 2018 is the top mass, as I seem to remember this was 2018 when doing something with epsK or B mixing).

### Some comments

With the new KS->pi mu nu and KS->pi e nu data, I removed the PDG values, since:
- For e the old PDG number was dominated by a 2006 KLOE measurement and the new result is actually a combination of that with the new data.
- For mu, the only measurement is the 2019 result from KLOE (the old number was, to quote the PDG, "not a measurement, calculate as 0.666 x BR(pi e nu)")

For the NA62 K+->pinunu, since the old number was just a preliminary result from a talk, I thought maybe in this case it does make sense to replace rather than add the new number.


MJKirk:
Oh, yes I forgot to fix the tests...
(Some of the errors are due to the latest version of numpy upgrading a depreciation warning to an error (see last bullet point in https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations).)

Actually for the beta decay, the SM prediction and the exp are going to be different with the new radiative corrections. Should I just increase the allowed error to make the test pass (since currently the allowed delta is just 2 or 3 times the experimental error), or actually alter the test with some reasonable calculation of the difference given the implied Vud values?

MJKirk:
Now it is just the numpy errors left, so I can rebase whenever you fix those on the main branch

peterstangl:
> Now it is just the numpy errors left, so I can rebase whenever you fix those on the main branch

I have fixed the numpy errors on the main branch, so you could do a rebase to rerun the tests.

MJKirk:
Forgot to comment - now ready to merge @peterstangl

MJKirk:
Closing, as the changes were merged through #218 and #219

PR #197, opened 2022-12-28
update eps_k parameter from arxiv 1911.06822

### LINKED ISSUE #182: epsilon_K from 1911.06822
As discussed with @MJKirk in https://github.com/flav-io/flavio/issues/172 it would be nice to change the implementation of epsilon_K to the more recent and more precise results from https://arxiv.org/abs/1911.06822.

amanmdesai:
Hi @peterstangl  I hope this PR Resolves #182. 
If not please let me know if something else needs to be updated.

Thanks

### DIFF
--- a/flavio/data/parameters_uncorrelated.yml
+++ b/flavio/data/parameters_uncorrelated.yml
@@ -322,7 +322,7 @@ BR(tau->rhonu): 25.2(3)e-2
 BR(Lambda(1520)->NKbar)_exp: 0.45 +- 0.01
 
 # Parameters needed for eps'/eps
-eps_K: 2.228(11)e-3
+eps_K: 2.16(18)e-3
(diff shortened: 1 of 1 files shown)


peterstangl:
Hi @amanmdesai,

the idea of #182 is to use the method of 1911.06822 to update the flavio implementation of the theory prediction of $\varepsilon_K$. The parameter with name `eps_K`, on the other hand, is used in the theory prediction of $\varepsilon'/\varepsilon$, corresponds to the measured value of $\varepsilon_K$ (see the [metadata](https://github.com/flav-io/flavio/blob/726bcb771891bdbd98d79d3732575752b02eaf1a/flavio/data/parameters_metadata.yml#L1090-L1092)) and should not be changed to the theory prediction of 1911.06822.

PR #203, opened 2023-02-28
Include new nuclear uncertainties in beta decays

MJKirk:
This adds the enlarged uncertainties from recent developments (see III.A.3 of Hardy:2020qwl for some background) to the super-allowed beta decays.
In Hardy:2020qwl (and everyone else I think), the new developments are applied as a extra fixed systematic uncertainty on the Ft values, as with the delta_R' uncertainties, and so I have included them here in a similar way.
I've split it into `delta_NS,A` and `delta_NS,E` (matching the naming in Hardy&Towner (note delta_NS,B is already included as a statistical uncertainty)).
`delta_NS,A` is the same for each nucleus, while `delta_NS,E` is given as a fixed number multiplied by the decay transition energy `Q_EC`. 
I have checked that my implementation:
1. Gives a very similar size for the combined NS uncertainties for each nucleus as using the method of Hardy&Towner
1. Gives the same size discrepancy between Vud from beta decays and Vus from kaon decays as found in the literature (such as 2208.11707) and doing a naive chi2 fit.

(I've marked this as a draft since right now my branch is based upon some of various changes from my other pull requests, and so may need rebasing anyway once those are merged, and at that point it will be clearer what this one actually changes. But at least now I am less likely to forget it.)

### DIFF
--- a/flavio/data/parameters_metadata.yml
+++ b/flavio/data/parameters_metadata.yml
@@ -1993,7 +1993,7 @@ deltaRp_n:
   description: Long-distance radiative correction to neutron beta decay
 delta_deltaRp_Z2:
   tex: $\delta(\delta R^\prime) / Z^2$
-  description: Uncertainty on the structure-indipendent radiative correction to beta decay divided by $Z^2$
+  description: Uncertainty on the structure-independent radiative correction to beta decay divided by $Z^2$
 delta_deltaNS 10C:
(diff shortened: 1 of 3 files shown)


MJKirk:
Okay @peterstangl, this PR is now ready for you to check out

peterstangl:
Thanks @MJKirk, this looks good!

PR #208, opened 2023-03-08
High-mass Drell-Yan

alekssmolkovic:
This PR implements the high-mass Drell-Yan differential cross-section predictions and experimental data, as described in https://arxiv.org/abs/2212.10497.

The neutral ($pp \to \ell \ell$) and charged current ($pp \to \ell \nu$) DY cross-section predictions are implemented in `flavio/physics/dileptons/ppll.py` and `flavio/physics/dileptons/pplnu.py`, respectively.

The `NNPDF40_nnlo_as_01180` PDF set is (down)loaded in `flavio/physics/dileptons/partondist.py`, relying on the [`parton` ](https://github.com/DavidMStraub/parton) package, which is now also added to the required packages in `setup.py`. The member id of the PDF replicas is defined to follow a discrete uniform distribution in `flavio/data/parameters_uncorrelated.yml` with the central value set to `0`. For this purpose the discrete uniform distribution is implemented as `DiscreteUniformDistribution` in `flavio/statistics/probability.py`.

### DIFF
--- a/flavio/physics/dileptons/ppll.py
+++ b/flavio/physics/dileptons/ppll.py
@@ -1,0 +1,231 @@
+from flavio.math.integrate import nintegrate
+from flavio.physics.zdecays.smeftew import gV_SM, gA_SM, _QN
+from flavio.physics.common import add_dict
+from flavio.classes import Observable, Prediction
+from flavio.physics import ckm as ckm_flavio
+from . import partondist
+import wilson
+from math import pi, sqrt
+import numpy as np
+
+def F_qqll_SM(q, Xq, l, Xl, s, par):
+    r"""SM $Z$ and $\gamma$ contribution to the $\bar q q\to \ell^+\ell^-$
+    amplitude."""
+    # parameters
+    Ql = -1
+    mZ = par['m_Z']
+    GammaZ = 1 / par['tau_Z']
+    s2w = par['s2w']
+    aEW = par['alpha_e']
+    e2 = 4 * pi * aEW
+    g2=e2/s2w * 0.9608 # correction factor from running g^2 to 1TeV
+    gp2=e2/(1-s2w) * 1.0272 # correction factor from running g'^2 to 1TeV
+    g_cw = sqrt(g2+gp2) # g over cw
+    s2w = gp2/(g2+gp2)
+    par = par.copy()
+    par['s2w'] = s2w
+    # SM couplings
+    Qq = _QN[q]['Q']
+    gVq = g_cw * gV_SM(q, par)
+    gVl = g_cw * gV_SM(l, par)
+    gAq = g_cw * gA_SM(q, par)
+    gAl = g_cw * gA_SM(l, par)
+    if Xq == 'L':
+        gZq = gVq + gAq
+    elif Xq == 'R':
+        gZq = gVq - gAq
+    if Xl == 'L':
+        gZl = gVl + gAl
+    elif Xl == 'R':
+        gZl = gVl - gAl
+    # SM contribution
+    return g2*s2w * Qq * Ql / s + gZq * gZl / (s - mZ**2 + 1j * mZ * GammaZ)
+
+
+def wceff_qqll_sm(s, par):
+    E = np.einsum('ij,kl->ijkl', np.eye(3), np.eye(3))
+    wc = {}
+    for Xl in 'LR':
+        for Xq in 'LR':
+            for q in 'ud':
+                wc['CV{}{}_e{}'.format(Xl, Xq, q)] = F_qqll_SM(q, Xq, 'e', Xl, s, par) * E
+    wc['CSRL_ed'] = wc['CSRR_ed'] = wc['CTRR_ed'] = np.zeros((3, 3, 3, 3))
+    wc['CSRL_eu'] = wc['CSRR_eu'] = wc['CTRR_eu'] = np.zeros((3, 3, 3, 3))
+    return wc
+
+
+def wceff_qqll_np(wc_obj, par, scale):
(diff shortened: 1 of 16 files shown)


flavio/physics/dileptons/test_pplnu.py
```diff
@@ -4,14 +4,18 @@
...
-from functools import lru_cache
+from flavio.config import config
 
 par = flavio.default_parameters.get_central_all()
 par2 = par.copy()
 par2['s2w'] = 0.22137905667111532  # correction factor to match the madgraph input scheme
 GeVtopb = 0.3894*10**9
+
+pdf_set_for_tests = 'NNPDF30_nnlo_as_0118'
+config['PDF set']['dileptons'] = pdf_set_for_tests
```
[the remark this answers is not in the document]
alekssmolkovic: You're right, these two lines are not necessary because of the import.

flavio/physics/dileptons/ppll.py
```diff
@@ -0,0 +1,231 @@
+from flavio.math.integrate import nintegrate
+from flavio.physics.zdecays.smeftew import gV_SM, gA_SM, _QN
+from flavio.physics.common import add_dict
+from flavio.classes import Observable, Prediction
+from flavio.physics import ckm as ckm_flavio
+from . import partondist
+import wilson
+from math import pi, sqrt
```
peterstangl, line 8: maybe we could use
```python
from numpy import pi, sqrt
```
such that the `sqrt` also works on arrays. And then we could replace all `np.pi` and `np.sqrt` by just `pi` and `sqrt`.

flavio/physics/dileptons/ppll.py
```diff
@@ -0,0 +1,231 @@
...
+    Parameters:
+    - `s`: hadronic center of mass energy in GeV$^2$
+    - `tau`: $\tau$, ratio of partonic to hadronic center of mass energy, dimensionless
+    - `l`: lepton flavour, should be 'e', 'mu', or 'tau'
+    - `Q2`: factorization scale squared in GeV$^2$
+    - `wc_eff`: SM + SMEFT amplitude for the process as dictionary with entries corresponding to different Wilson coefficients
+    - `par`: parameter dictionary
+    """
+    sigma = 0
+    member = int(par['NNPDF40_nnlo_as_01180 member'])
```
peterstangl, line 149: I'm not sure if we should hard-code something related to the PDF set here since in principle the PDF set could be changed using `flavio.config`. But maybe the PDF member numbers are actually independent of the explicit PDF set used and then we could change `'NNPDF40_nnlo_as_01180 member'` to `'PDF member'`. This we would need to check.

flavio/physics/dileptons/pplnu.py
```diff
@@ -0,0 +1,196 @@
+from flavio.math.integrate import nintegrate
+from flavio.physics.common import add_dict
+from flavio.physics import ckm as ckm_flavio
+from flavio.classes import Observable, Prediction
+from . import partondist
+import wilson
+from math import pi
```
peterstangl, line 7: Same comment as for `ppll.py`: we could use
```python
from np import sqrt, pi
```
and replace all `np.pi` and `np.sqrt` by just `pi` and `sqrt`

flavio/physics/dileptons/ppll.py
```diff
@@ -0,0 +1,231 @@
...
+            i1 = pdf_flavor[q1]
+            i2 = pdf_flavor[q2]
+            sh = tau * s
+            y = sigma_qqll_partonic(sh, q1, q2, l, wc_eff)
+            if y != 0:  # save time if the contribution is anyway zero
+                sigma += 2 * plumi.L(i1, -i2, tau) * y
+    return sigma
+
+
+def sigma_qqll_int(s, q2min, q2max, l, Q2, wc_obj, par, scale, newphys=True):
```
peterstangl, line 166: I wonder if we should use arguments `qmin` and `qmax` instead of `q2min` and `q2max` since this is also what we use for the observable.

flavio/physics/dileptons/ppll.py
alekssmolkovic: Thanks, I have fixed this in the latest comit.

flavio/physics/dileptons/ppll.py
alekssmolkovic: A potential issue with that is that in `parameters_uncorrelated.yml` one has to provide the lowest and highest value to `discrete_uniform`, and the number of pdf members might not always be 100. So in that sense, `NNPDF40_nnlo_as_01180 member` is more descriptive (although it can be used for any pdf set with 100+1 members).

flavio/physics/dileptons/ppll.py
alekssmolkovic: I agree, I have changed to `qmin` and `qmax` everywhere in `ppll.py`, as well as modified `test_ppll.py` to reflect the change.

flavio/physics/dileptons/ppll.py
peterstangl: Exactly. So the question would be if it's always 100+1 members. But even if it is not, one could just update the parameter `PDF member` using a different `discrete_uniform` distribution. So also in that case it might make sense to not have the name of the PDF set in the parameter name, no?

alekssmolkovic: Thanks for resolving this in the latest commit!

peterstangl:
OK, this PR looks good and I think it could be merged. Now we just have to wait for the new version of `parton` to be released, which includes PR https://github.com/DavidMStraub/parton/pull/8.

peterstangl:
Thank you @DavidMStraub! I think this PR can be merged now.

Issue #209, opened 2023-03-14
Gamma distributions


peterstangl:
The current implementation of probability distributions related to the Gamma distribution might be confusing for the following reasons:

- `GeneralGammaDistributionPositive` is not a generalization of `GammaDistributionPositive` since instead of the arguments `a`, `loc`, and `scale`, it has arguments `scale_factor`, `counts_total`, `counts_signal`.
- No equivalent of `GeneralGammaDistributionPositive` with vanishing background uncertainty exists with arguments `scale_factor`, `counts_total`, `counts_signal`.
- The argument `background_variance` used in `GeneralGammaDistributionPositive` and `GeneralGammaUpperLimit` is not the variance but the standard deviation.

peterstangl:
All the above points have been addressed in PR #210.

PR #210, opened 2023-03-14
Gamma distributions

peterstangl:
This PR is a first step to address issue #209.

- The plan is to make `GeneralGammaDistributionPositive` a generalization of `GammaDistributionPositive` with arguments `a`, `loc`, and `scale`. However, this would break backward compatibility. Therefore, the temporary class `GeneralGammaDistributionPositiveTemporary` is introduced that fulfills this purpose. It will be renamed to `GeneralGammaDistributionPositive` in the future.
- The new class `GeneralGammaCountingProcess` is introduced. It takes arguments `scale_factor`, `counts_total`, `counts_signal`, and `background_std` and replaces the functionality of the current class `GeneralGammaDistributionPositive` (which is kept for backward compatibility for the moment, but `GeneralGammaDistributionPositiveTemporary` will be renamed to `GeneralGammaDistributionPositive` in the future)
- The new class `GammaCountingProcess` is introduced as an equivalent of the current `GeneralGammaCountingProcess` with vanishing background uncertainty. It takes arguments `scale_factor`, `counts_total`, and `counts_signal`.

The inheritance is now as follows:
- `GammaDistributionPositive` &rightarrow; `GammaCountingProcess` &rightarrow; `GammaUpperLimit`
- `GeneralGammaDistributionPositive` &rightarrow; `GeneralGammaCountingProcess` &rightarrow; `GeneralGammaUpperLimit`

### DIFF
--- a/flavio/data/measurements.yml
+++ b/flavio/data/measurements.yml
@@ -5219,126 +5219,126 @@ Belle B->hnunu SL 2017:
   values:
     # Tables I and II
     # counts_total is the square of the statistical error on the signal yield
-    # background_variance is the systematic error on the signal yield
+    # background_std is the systematic error on the signal yield
     BR(B+->Knunu):
       distribution: general_gamma_upper_limit
       limit: 1.9e-5
       confidence_level: 0.9
       counts_total: 83
       counts_signal: 17.7
-      background_variance: 3.4
+      background_std: 3.4
     BR(B0->Knunu):
       distribution: general_gamma_upper_limit
       limit: 2.6e-5
       confidence_level: 0.9
       counts_total: 18
       counts_signal: 0.6
-      background_variance: 1.4
+      background_std: 1.4
     BR(B+->K*nunu):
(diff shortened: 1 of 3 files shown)


DavidMStraub:
I would suggest avoiding `GeneralGammaDistributionPositiveTemporary` because even if you have deprecation warnings, it will not be possible for users to migrate to the new class without introducing another breaking change in the future when you rename it back. You can't avoid a breaking change, so just announce it and make a clear cut; ideally, make all arguments of the new class's constructor keyword only, to make sure non-migrated code will raise rather than silently calculating a wrong distribution.

peterstangl:
I think since `GeneralGammaDistributionPositive` was not used directly in flavio so far (only as base class of `GeneralGammaUpperLimit`, which is used for some measurements), and it probably is also not used much outside of flavio, it should be safe to break backward compatibility there. In any case, I made the new arguments of  `GeneralGammaDistributionPositive` keyword only such that using the old interface will raise an error.

PR #211, opened 2023-03-24
Implementation of b>dll observables

alekssmolkovic:
This PR adds three new observables with an underlying $b \to d \ell \ell$ transition:

- Differential branching ratio of $B \to \pi \mu \mu$ (LHCb)
- Branching ratio of $B_s \to K^{*0} \mu \mu$ (LHCb)
- Branching ratio of $B\to \pi e e$ (Belle)

In the case of $B_s \to K^{*0} \mu \mu$, the procedure outlined in Appendix B of [Bause:2022rrs](https://arxiv.org/pdf/2209.04457.pdf) is followed to account for the uncertainty which arises from the interpretation of LHCb data. This is implemented with the `delta_BsKstarmumu` parameter. 

Subleading corrections are currently neglected due to large experimental uncertainties.


flavio/data/measurements.yml
```diff
@@ -22300,3 +22300,32 @@ CMS pp->munu 2022:
       counts_total: 0.0
       distribution: gamma_counting_process
       scale_factor: 15181.392014563447
+
+Belle B+->pi+ee 2008:
+  experiment: Belle
+  inspire: Belle:2008tjs
+  values:
+    BR_Belle(B+->piee): < 8e-8 @ 90% CL
```
peterstangl, line 22308: Maybe it's better to implement the measured value for the BR instead of the upper limit if this measurement is then used for a likelihood with a Gaussian approximation.

flavio/data/parameters_metadata.yml
```diff
@@ -2010,3 +2010,10 @@ Lambda->p g_1(0):
 PDFmembers avg=0 replicas=1-100:
   tex: $\text{mem}_{\text{PDF}}$
   description: member ID of PDF set where mem=0 is average on replicas and mem=1-100 are PDF replicas
+
+# Parameter for Bs->K*0mumu uncertainty inside resonant regions
+# 2209.04457, appendix B, Eq. B7, we save a relative uncertainty
+delta_BsKstarmumu:
+  tex: $\delta_{\text{BsKstarmumu}}$
+  tex: $\delta_{B_s^0 \to \bar{K}^{\ast 0} \mu \mu}$
```
peterstangl, line 2018: why are there two `tex` entries? I think one should be removed

alekssmolkovic: Thanks, I have removed the redundant tex entry.

flavio/data/measurements.yml
alekssmolkovic: Good point, in the latest commit I have changed to the measured value reported in the reference.

peterstangl:
Thanks, this looks good!

PR #218, opened 2023-04-21
CAA related updates

MJKirk:
This is just the CAA related measurement updates from #196. 
This is most of the new inputs I made use of in my paper on the CAA (https://arxiv.org/abs/2212.06862)
For KS->pi mu nu and KS->pi e nu, I removed the PDG values from the group and added them separately, since:
1. For e the old PDG number was dominated by a 2006 KLOE measurement and the new result is actually a combination of that with the new data.
1. For mu, the only measurement is the 2019 result from KLOE (the old number was, to quote the PDG, "not a measurement, calculate as 0.666 x BR(pi e nu)")


peterstangl:
Thank you very much @MJKirk!

PR #219, opened 2023-04-21
Misc updates 2023

MJKirk:
Updating the PDG year used for the masses, and the value of gamma (these are the other updates from #196).


peterstangl:
Thank you @MJKirk!

PR #246, opened 2023-12-14
change CKM delta->gamma and add beta-gamma implementation

peterstangl:
This PR implements the "beta-gamma" parameterization of the CKM matrix based on the update of the `ckmutil` package in PR https://github.com/DavidMStraub/ckmutil/pull/4.

Both the "beta-gamma" and "tree" CKM parameterizations use the Unitarity Triangle (UT) angle $\gamma$ to determine the CKM phase $\delta$. Since $\delta=\gamma$ is a very accurate approximation, so far $\delta$ was set to $\gamma$ everywhere in `flavio` and the parameter corresponding to the angle $\gamma$ has been called `delta`. This is changed by the present PR, which now distinguishes $\gamma$ and $\delta$ and uses the name `gamma` for the UT angle $\gamma$, while keeping the name `delta` for the CKM phase $\delta$, which is used in the standard CKM parameterization.
This change is motivated by the implementation of the exact relation between $\gamma$ and $\delta$ in the `ckmutil` package in PR https://github.com/DavidMStraub/ckmutil/pull/4.


DavidMStraub:
Silly question, why do tests pass if the ckmutil PR is not yet merged?

peterstangl:
> Silly question, why do tests pass if the ckmutil PR is not yet merged?

- The change of `delta` -> `gamma` is just renaming the parameter but doesn't need the ckmutil PR
- The new implementation of the beta-gamma scheme is not used as the default one at the moment and there is no test in flavio that explicitly tests this input scheme

flavio/physics/ckm.py
```diff
@@ -20,16 +20,20 @@ def ckm_standard(t12, t13, t23, delta):
...
 
 @lru_cache(maxsize=2)
-def ckm_tree(Vus, Vub, Vcb, delta):
-    return ckmutil.ckm.ckm_tree(Vus, Vub, Vcb, delta)
+def ckm_tree(Vus, Vub, Vcb, gamma):
+    return ckmutil.ckm.ckm_tree(Vus, Vub, Vcb, gamma)
+
+@lru_cache(maxsize=2)
+def ckm_beta_gamma(Vus, Vcb, beta, gamma):
+    return ckmutil.ckm.ckm_beta_gamma(Vus, Vcb, beta, gamma)
```
peterstangl, line 36: @DavidMStraub I think this is the only line in the code that requires the new PR in `ckmutil` and it is never called in any test.

DavidMStraub:
ckmutil 1.1.0 is on PyPI.

PR #251, opened 2024-01-08
fix `diff_plot_th_err` function to correctly utilize argument `N`

girishky:
Addressed a small issue in the function `diff_plot_th_err`, where the `N` parameter wasn't accessed for computing uncertainties. Previously, the functions `flavio.sm_uncertainty` and `flavio.sm_uncertainty` within `diff_plot_th_err` were not utilizing the provided `N` value, resulting in the default value (`N`=100) being used for all uncertainty calculations.


DavidMStraub:
Oops, good catch

peterstangl:
Thanks for this PR @girishky!

PR #252, opened 2024-01-15
use beta gamma CKM scheme as default

peterstangl:
This PR changes the default CKM scheme to the `beta gamma` scheme, which, compared to the `Tree` scheme, uses the UT angle $\beta$ instead of $|V_{ub}|$. The advantage is that this input scheme is unaffected by the tension between inclusive and exclusive $|V_{ub}|$ determinations (and the tensions between different exclusive determinations). Furthermore, with the most recent determination of $\beta$, the precision of the CKM elements is increased. The predicted value of $|V_{ub}|$ in this scheme with the current `flavio` inputs is
$$|V_{ub}|=(3.72\pm0.09)\times 10^{-3}\\,,$$
which has an error nearly two times smaller than the exclusive $V_{ub}$ determined from $B\to\pi\ell\nu$.

### DIFF
--- a/flavio/physics/test_ckm.py
+++ b/flavio/physics/test_ckm.py
@@ -45,6 +45,7 @@ class TestCKM(unittest.TestCase):
     par_s = dict(t12=t12,t13=t13,t23=t23,delta=delta)
     par_w = dict(laC=laC,A=A,rhobar=rhobar,etabar=etabar)
     par_t = dict(Vus=Vus,Vub=Vub,Vcb=Vcb,gamma=gamma)
+    par_b = dict(Vus=Vus,Vcb=Vcb,beta=beta,gamma=gamma)
 
     def test_ckm_parametrizations(self):
         np.testing.assert_almost_equal(self.v_t/self.v_s, np.ones((3,3)), decimal=5)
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Is that beta from B to psi KS or something else?

peterstangl:
It's mostly $B\to\psi K_S$ but also some other charmonium channels listed here: https://hflav-eos.web.cern.ch/hflav-eos/triangle/latest/#sin2b

DavidMStraub:
But then it's no longer a tree-level determination, so not appropriate to use it in presence of dim-6 NP, or not?

peterstangl:
But shouldn't the question be whether there might be a potential dim-6 NP contribution affecting the extraction of the CKM parameter, rather than whether the process is tree or loop level? Of course, in general, loop-level processes might be more affected by dim-6 NP than tree-level ones, but there is no real guarantee of that, right? And at the moment there are tensions in tree-level $b\to u\ell\nu$ data that could be interpreted in terms of dim-6 NP (e.g. https://arxiv.org/abs/2302.05268).

So, in the end, don't we always have to choose which dim-6 NP contributions we assume to be absent when extracting the CKM elements as used in the current implementation in flavio? And I'm not sure the absence of NP in tree-level $b\to u\ell\nu$ is necessarily better motivated than its absence in loop-level $B$ mixing.

In the end, could it make sense to implement the CKM extraction in the presence of dim-6 NP that we have in `smelli` in `flavio` to avoid this problem no matter whether tree or loop level observables are used?

PR #255, opened 2024-07-16
fix issue in Constraints.add_constraint

peterstangl:


DavidMStraub:
Why aren't you keeping the existing behaviour as default to preserve backward compatibility?

peterstangl:
I think that's what I'm doing now with this PR using the argument `parameter_object=False`. I introduced an issue in a previous commit in PR https://github.com/flav-io/flavio/pull/247 where I actually broke backward compatibility. The present PR should restore it.

DavidMStraub:
Perhaps parameter_object might benefit from a different name that fits to the fact that it's a boolean, not a parameter object

peterstangl:
Maybe something like `is_parameter_constraint`? :thinking: The main point is just that this should only be applied to parameter constraints and not to measurement constraints.

PR #267, opened 2025-03-03
fix typo in LHCb RD* 2023

AlexandreCarvunis:
Update of LHCb RD* 2023. The initial commit corresponds to the first published result (https://arxiv.org/abs/2305.01463v1) which was updated in may 2024 (https://arxiv.org/abs/2305.01463v2).

This new value is a combination including the result of LHCb RD* 2017 (1708.08856)


peterstangl:
Great, thank you!

PR #269, opened 2025-05-05
Delete old deploy script

DavidMStraub:
Just saw this - this is no longer needed, we deploy via workflow.


peterstangl:
Great, thank you!

PR #270, opened 2025-05-05
Use pyproject.toml

DavidMStraub:
Only change: switched test optional dependency from nose2 to pytest.


peterstangl:
Thanks! That's a good idea to update to pyproject.toml :smile:

I know Python 3.8 has reached EOL last October. But in principle flavio would still work with 3.7, no? So is it really necessary to increase the min Python version as long as there is no code or dependency that require a newer version?

DavidMStraub:
The reason I thought support is officially dropped is because <3.9 is not tested on the CI. I think without testing in the CI, it is not really "supported". I suggest following [SPEC 0](https://scientific-python.org/specs/spec-0000/) and dropping support. But I agree, it's a decision independent of the build system switch.

peterstangl:
OK, yes that's a good point! Without CI it would be very difficult to know if a new commit would actually work with 3.7 or 3.8. And then it's probably better to not support these old versions!

PR #276, opened 2025-09-17
migrate from pkg_resources to importlib

### LINKED ISSUE #271: pkg_resources slated for removal
Importing `flavio` with the latest `setuptools` (80.9.0) installed raises a warning stating

```
UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. 
The pkg_resources package is slated for removal as early as 2025-11-30.
Refrain from using this package or pin to Setuptools<81.
```

`pkg_resources` is currently used in two places: https://github.com/flav-io/flavio/blob/b70050adef5bb286049aa26414f7f406935d36d8/flavio/physics/bdecays/bvll/qcdf_interpolate.py#L17 and https://github.com/flav-io/flavio/blob/b70050adef5bb286049aa26414f7f406935d36d8/flavio/physics/bdecays/wilsoncoefficients.py#L15

According to the warning, it would be good to migrate these soon to [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources)

girishky:
This fixes #271

### DIFF
--- a/flavio/physics/bdecays/bvll/qcdf_interpolate.py
+++ b/flavio/physics/bdecays/bvll/qcdf_interpolate.py
@@ -14,14 +14,17 @@ precision.
 import flavio
 import numpy as np
 import pkgutil
-import pkg_resources
+import importlib.resources as resources
 import scipy.interpolate
 import math
 import warnings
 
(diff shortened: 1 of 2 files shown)


DavidMStraub:
Looks good to me!

PR #294, opened 2026-06-15
Update a_mu SM and exp

alekssmolkovic:
This PR updates the SM prediction of `a_mu` to the one reported [here](https://inspirehep.net/literature/2925594) and the FNAL combination in measurements to the one reported [here](https://inspirehep.net/literature/2928642). The `FNAL g-2 combination` measurement is used in `smelli`.

### DIFF
--- a/flavio/data/measurements.yml
+++ b/flavio/data/measurements.yml
@@ -8231,6 +8231,12 @@ FNAL g-2:
 FNAL g-2 combination:
   experiment: Muon g-2
   values:
+    a_mu: 116592061(41) e-11
+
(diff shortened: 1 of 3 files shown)


peterstangl:
Thanks!

For the measurement, I think we should add a new entry that contains the year in the name, e.g. `FNAL g-2 combination 2025` instead of modifying the old measurement, which would break backwards compatibility. Otherwise, old results in existing code could not be reproduced anymore. Please also include the `inspire:` field in the new entry.

I think the tests also have to be updated.

